Issue #32, opened 2021-03-02
cdrom code


sezero:
We have stubs for cdrom code in sdl12-compat.

Should we import the code from SDL-1.2? If we ever do that,
would there be licence issues (can any of the authors still be
reached, etc)?

If we'll keep the stubs, close this.

icculus:
I was looking at this last night, because at some point the ScummVM team told me not having CD support was a deal breaker, but we can't just paste that code in, and in some cases, it wouldn't be helpful (for example, Apple doesn't sell Macs with CD-ROM drives anymore, and for years before, the audio cable wasn't hooked up, so you couldn't just tell an audio CD to play in one...iTunes would play CDs by reading the raw data from the disc and playing it through CoreAudio, instead of just sending a command to the drive like these SDL APIs expect).

Also it feels weird to think that implementing this properly means that half of sdl12-compat would be CD-ROM code for various platforms.  :)

Let's leave this open for now, but my suspicion is that modern hardware and a lack of readily-available discs is going to be the real problem in the future.

Part of me thinks it could be cool if we could configure sdl12-compat to pretend a directory full of MP3s is an audio CD, though.

sezero:
Side note: libcdio could be used for cd-play, but it's GPLv3, not LGPL:
https://www.gnu.org/software/libcdio/

icculus:
> Part of me thinks it could be cool if we could configure sdl12-compat to pretend a directory full of MP3s is an audio CD, though.

I did it.  :)   It's WIP, but almost done.

icculus:
okay, I finished this work over the weekend.

Tested with quakespasm and the original Quake 1 soundtrack ripped to mp3, and it seems to work pretty well.

The only issue I see at the moment is that it doesn't have a way to specify data tracks, so for Quake to get the right music (which starts on track 2, past the game data at track 1), you have to copy a valid mp3 file to 0.mp3 to make sdl12-compat report tracks at the right place. This obviously should be fixed still, so I'm reopening this bug for now.

icculus:
Also we should probably start with 1.mp3 instead of 0.mp3 so it matches the track number, and maybe call these things "track1.mp3" instead.

sezero:
Are the static linkage issues fixed? (Haven't tested yet.)

> Also we should probably start with 1.mp3 instead of 0.mp3 so it matches the track number, and maybe call these things "track1.mp3" instead.

[...]

PR #36, opened 2021-03-08
cdrom: Initial shot at fake CD-ROM support.

icculus:
This makes a directory of MP3 files look like an audio CD to the 1.2 app,
so something like (for example) Quake 1 can rip the retail disc's audio
to the filesystem, and an SDL 1.2 version of the game will still believe the
soundtrack is playing from disc.

This requires us to share the audio device with the 1.2 SDL_OpenAudio
interface, which presents some other tapdancing, but it appears to be a
feasible approach.

This is work in progress; it compiles but there are small pieces missing
and it hasn't been debugged at all. Do not merge yet!

This is in reference to Issue #32.


sezero:
I am not very happy with the general idea. An extra effort is needed from the user who possibly isn't aware of it. But it's just me.

icculus:
Yeah, it would need effort from the user, but my concerns are that hardware and discs aren't going to be available in the future--as they barely are now--in which case offering real CD support isn't helpful.

We _could_ offer real hardware options and _also_ supply a virtual disc like this, but the hardware code is big and messy. But that's not a reason to avoid that approach if it's the right one.

For the Mac (if not every platform), we're definitely going to have to read the physical disc and share the audio device, so this code isn't a total loss even if we go that direction.

I'm going to work on some other issues for awhile and think on it further. Feedback is definitely welcome.

slouken:
I think if we have games that people are actually playing that rely on that functionality, this is a great way to provide compatibility so they continue to run. I think in that case, people will go to the effort, because that's the way they can run the games in modern environments.

If we don't think there's a real-world use-case, then this is probably not worth spending time on.

sezero:
In any case, here are a few build results:

- Windows -nostdlib builds are an issue:
  I got several `undefined reference to '___chkstk'` from x64 builds.
  From x86 builds, additional undefined references to `___udivdi3` and
  possibly to `_alloca`. I didn't follow the code to find the reason.
  For mingw, linking with -lgcc solves it:
```diff
diff --git a/src/Makefile.mingw b/src/Makefile.mingw
index 085ffcb..b509c69 100644
--- a/src/Makefile.mingw
+++ b/src/Makefile.mingw
@@ -17,7 +17,7 @@ CPPFLAGS = -DDLL_EXPORT -DNDEBUG
 
 CFLAGS  = -O3 -Wall
 LDFLAGS = -nostdlib -shared -Wl,--no-undefined -Wl,--enable-auto-image-base -Wl,--out-implib,$(LIB)
-LDLIBS  = -lkernel32 -luser32
+LDLIBS  = -lkernel32 -luser32 -static-libgcc -lgcc
 
 LIB = libSDL.dll.a
 DLL = SDL.dll
```
  For MSVC, don't know what to do.

- The following patch addresses C90 issues and adds missing return values:
```diff
diff -u a/src/SDL12_compat.c b/src/SDL12_compat.c
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -4794,7 +4794,6 @@
 static void
 InitializeCDSubsystem(void)
 {
-    int trackcount = 0;
     const char *cdpath;
     FIXME("Is subsystem init reference counted in SDL 1.2?");  /* it is in SDL2, but I don't know for 1.2. */
     if (CDRomInit) {
@@ -5140,6 +5139,7 @@ SDL_CDPause(SDL12_CD *cdrom)
         cdrom->status = audio_cbdata->cdrom_status;
     }
     SDL20_UnlockAudio();
+    return 0;
 }
 
 DECLSPEC int SDLCALL
@@ -5159,6 +5159,7 @@ SDL_CDResume(SDL12_CD *cdrom)
         cdrom->status = audio_cbdata->cdrom_status;
     }
     SDL20_UnlockAudio();
+    return 0;
 }
 
 
@@ -5184,6 +5185,7 @@ SDL_CDStop(SDL12_CD *cdrom)
     if (oldrw) {
         SDL20_RWclose(oldrw);
     }
+    return 0;
 }
 
 DECLSPEC int SDLCALL
@@ -5410,11 +5412,11 @@ CloseSDL2AudioDevice(void)
 static void
 CreateAudioStreams(void)
 {
-    FIXME("Check for errors in here");  /* ...but what are you supposed to do if it fails? Halt all audio? (Probably.) */
     AudioCallbackWrapperData *data = audio_cbdata;
     const SDL_AudioSpec *have = &data->device_format;
     size_t mixbuflen = have->size;
     void *ptr;
+    FIXME("Check for errors in here");  /* ...but what are you supposed to do if it fails? Halt all audio? (Probably.) */
 
     if (data->app_callback_opened) {
         const SDL_AudioSpec *want = &data->app_callback_format;
```

- The following one silences an annoying watcom warning:
```diff
diff -u a/src/dr_mp3.h b/src/dr_mp3.h
--- a/src/dr_mp3.h
+++ b/src/dr_mp3.h
@@ -4177,7 +4177,7 @@ static float* drmp3__full_read_and_close
 
             oldFramesBufferSize = framesCapacity * pMP3->channels * sizeof(float);
             newFramesBufferSize = newFramesCap   * pMP3->channels * sizeof(float);
-            if (newFramesBufferSize > DRMP3_SIZE_MAX) {
+            if (newFramesBufferSize > (drmp3_uint64)DRMP3_SIZE_MAX) {
                 break;
             }
 
@@ -4244,7 +4244,7 @@ static drmp3_int16* drmp3__full_read_and
 
             oldFramesBufferSize = framesCapacity * pMP3->channels * sizeof(drmp3_int16);
             newFramesBufferSize = newFramesCap   * pMP3->channels * sizeof(drmp3_int16);
-            if (newFramesBufferSize > DRMP3_SIZE_MAX) {
+            if (newFramesBufferSize > (drmp3_uint64)DRMP3_SIZE_MAX) {
                 break;
             }
 
```

sezero:
> For MSVC, don't know what to do.

Hm, MSVC seems to provide chkstk.obj, can try adding that to linkage later..

sezero:
> > For MSVC, don't know what to do.
> 
> Hm, MSVC seems to provide chkstk.obj, can try adding that to linkage later..

For x64 builds, adding chkstk.obj to LDLIBS was enough. But for x86, we have
more missing syms due to 64 bit math:
`__allmul, __aulldiv, __dtoui3, __dtoul3, __ftoui3, __ultod3`

rtissera:
Jumping in the conversation, what about having : 
- An environment variable like SDL_CDROM_IMAGE=file.iso / file.cue
- An ISO9660 or BIN/CUE reader
?

icculus:
My goal is to not have half the project be CD-ROM support code, but I'm not against whatever works.

sezero:
I tok the liberty of rebasing the patch to current master, with a few build fixes (Windows (msvc) issues remain).

icculus:
Took the latest from here and pushed a revised version to main: https://github.com/libsdl-org/sdl12-compat/commit/e5a32aa721c54874ee074c4edac18c746518d5ff

So this PR can be closed.

PR #41, opened 2021-04-11
Basic unicode input support

sulix:
This change adds unicode input support via the `SDL_TEXTINPUT` events.

These are turned into a number of new `KEYDOWN` events, one per input character. These generated events all have a sym of `SDLK_UNKNOWN,` so shouldn't register as spurious keyboard inputs. Normal key messages (i.e., those which did not originate with an `SDL_TEXTINPUT` event) now have a `unicode` value of 0 all the time.

It's not perfect yet (see below), but seems to be an improvement over just using the `keysym.sym` value, which didn't support things like using Shift to get uppercase letters.

Note that this also includes a basic UTF-8 decoder, which depends on  `SDL_MostSignificantBitIndex32()`. Since it's force-inlined, 

It also doesn't yet support codepoints greater than 0xFFFF, as we don't yet generate surrogate pairs.

I've tested this against a couple of different games (Defcon and Psychonauts), and input works. This also includes using the compose key to input accented characters and IBUS for more complicated input (including having several characters being input as part of the same event). While CJK characters weren't working (likely due to the UTF-16 surrogate pair issue), and there were issues with some games which have limited font support or similar, it seems to work a little better than I expected so far.

### DIFF
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -85,6 +85,67 @@
 #define SDL20_zerop(x) SDL20_memset((x), 0, sizeof(*(x)))
 #define SDL_ReportAssertion SDL20_ReportAssertion
 
+/* From SDL2.0's SDL_bits.h: a force-inlined function. */
+#if defined(__WATCOMC__) && defined(__386__)
+extern _inline int _SDL20_bsr_watcom (Uint32);
+#pragma aux _SDL20_bsr_watcom = \
+    "bsr eax, eax" \
+    parm [eax] nomemory \
+    value [eax] \
+    modify exact [eax] nomemory;
+#endif
+
+SDL_FORCE_INLINE int
+SDL20_MostSignificantBitIndex32(Uint32 x)
+{
+#if defined(__GNUC__) && (__GNUC__ >= 4 || (__GNUC__ == 3 && __GNUC_MINOR__ >= 4))
+    /* Count Leading Zeroes builtin in GCC.
+     * http://gcc.gnu.org/onlinedocs/gcc-4.3.4/gcc/Other-Builtins.html
+     */
+    if (x == 0) {
+        return -1;
+    }
+    return 31 - __builtin_clz(x);
+#elif defined(__WATCOMC__) && defined(__386__)
+    if (x == 0) {
+        return -1;
+    }
+    return _SDL20_bsr_watcom(x);
+#elif defined(_MSC_VER)
+    unsigned long index;
+    if (_BitScanReverse(&index, x)) {
+        return index;
+    }
+    return -1;
+#else
+    /* Based off of Bit Twiddling Hacks by Sean Eron Anderson
+     * <seander@cs.stanford.edu>, released in the public domain.
+     * http://graphics.stanford.edu/~seander/bithacks.html#IntegerLog
+     */
+    const Uint32 b[] = {0x2, 0xC, 0xF0, 0xFF00, 0xFFFF0000};
+    const int    S[] = {1, 2, 4, 8, 16};
+
+    int msbIndex = 0;
+    int i;
+
+    if (x == 0) {
+        return -1;
+    }
+
+    for (i = 4; i >= 0; i--)
+    {
+        if (x & b[i])
+        {
+            x >>= S[i];
+            msbIndex |= S[i];
+        }
+    }
+
+    return msbIndex;
+#endif
+}
+
+
 #define SDL12_DEFAULT_REPEAT_DELAY 500
 #define SDL12_DEFAULT_REPEAT_INTERVAL 30
 
@@ -471,6 +532,9 @@ typedef struct SDL12_keysym
     SDL12Mod mod;
     Uint16 unicode;
 } SDL12_keysym;
+
+#define SDL12_RELEASED 0
+#define SDL12_PRESSED  1
 
 typedef enum
 {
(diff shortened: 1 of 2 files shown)


src/SDL12_compat.c
```diff
@@ -85,6 +85,67 @@
 #define SDL20_zerop(x) SDL20_memset((x), 0, sizeof(*(x)))
 #define SDL_ReportAssertion SDL20_ReportAssertion
 
+/* From SDL2.0's SDL_bits.h: a force-inlined function. */
+#if defined(__WATCOMC__) && defined(__386__)
+extern _inline int _SDL20_bsr_watcom (Uint32);
+#pragma aux _SDL_bsr_watcom = \
```
sezero, line 91: This needs renaming to `_SDL20_bsr_watcom`

sezero:
SDL-1.2 build of quakespasm doesn't work OK - E.g.: use `Shift`+`Esc` to bring down
the game's console, try typing something, you can't.

src/SDL12_compat.c
sulix: Whoops! Fixed now.

sulix:
Okay: the quakespasm issue seems to be that it needed the "state" to be SDL_PRESSED on keyboard events, or it discarded them. (This was actually a little bit confusing, as the in_debugkeys variable in quakespasm seems to check this, but report it as though it were the event type. Bizarre.)

In any case, I've updated this, and quakespasm's console input works for me now. I did add SDL12_ variants of the PRESSED and RELEASED constants — they're the same values as the SDL 2 ones, and the rest of the keyboard event code relies on that, so it'd be possible to just re-use the SDL 2 ones if people would prefer that.

sezero:
This works OK for me, now thanks.

icculus:
Just making a note here: this is clearly a major improvement (thank you!!), but in theory we're going to get two KEYDOWN events for most physical key presses now, right? One with a `unicode` value of zero followed by a second event with a correct value?

I'm thinking we should hold the last key press event until a TEXTINPUT event comes through if SDL_EnableUNICODE(1) has been called.

- if it matches, send on a single KEYDOWN event with the correct `unicode` field set.
- if it doesn't match, send on the KEYDOWN event with `unicode` set to zero
- if another KEYDOWN comes in or we've held the keypress for ...50...? milliseconds, send it on with `unicode` set to zero.

This might not be the right approach either; it feels a little fragile, but that might be unavoidable.

In practice, maybe nothing cares about this, but I could see a game that ignores the unicode field and built their own text input code on keydown events getting double characters on each keypress (in SDL-1.2, key repeat was signified by multiple KEYDOWN events in a row without a KEYUP).

Should I open a new issue for this?

Also, as a second issue probably worth opening: SDL-1.2's unicode support is kind of broken: I suspect we should drop all but the first character in a TEXTINPUT event; I don't think 1.2 would send multiple events for longer-than-one-character responses, as it didn't have a concept of IME...if it couldn't apply a character to a specific press, you were out of luck.

That being said, I don't want to sound negative by pointing these things out; this is a huge step forward in compatibility for sdl12-compat, and I'm grateful for the work.

icculus:
> This might not be the right approach either; it feels a little fragile, but that might be unavoidable.

Thinking further: SDL-1.2 unicode support was so bad that it might have only been useful for extracting latin1 characters; in which case, perhaps we should just have a simple mapping from keycode to unicode value and ignore SDL_TEXTINPUT altogether.

slouken:
> > This might not be the right approach either; it feels a little fragile, but that might be unavoidable.
> 
> Thinking further: SDL-1.2 unicode support was so bad that it might have only been useful for extracting latin1 characters; in which case, perhaps we should just have a simple mapping from keycode to unicode value and ignore SDL_TEXTINPUT altogether.

I like the idea of separate key events for scancodes and unicode, but I also worry about games that interpret both. Maybe we can leave the current behavior and add simple scancode mapping via a hint if we find it's a problem in practice?

sulix:
So, I looked into how the unicode events in 1.2 actually behave, and it seems to be pretty different depending on platform/configuration (particularly on X11, where there are several different implementations). Of these the [XIM backend](https://github.com/libsdl-org/SDL-1.2/blob/main/src/video/x11/SDL_x11events.c#L702) behaves pretty similarly to this change — several SDL_KEYDOWN events with the sym as SDLK_UNKNOWN. It does use the last such event as the actual KEYDOWN event (with the correct sym set), but we'd need to do some sort of waiting and merging of events as you mentioned to manage that.

[Another part of the X11 implementation](https://github.com/libsdl-org/SDL-1.2/blob/main/src/video/x11/SDL_x11events.c#L774) does discard all but the first character produced though, so it's definitely inconsistent.

Similarly, I considered having a hint to fallback to the previous implementation of using the SDLK_ values (and maybe doing our own mappings for Shift/Caps Lock/etc), but thought I'd see how well this worked first.

Long story short, I _think_ that anything broken by this is probably broken on some real SDL 1.2 configurations as well (albeit possibly obscure ones), so it seemed pretty safe to go with this as-is and fix it if we find something which actually breaks.

sezero:
> Just making a note here: this is clearly a major improvement (thank you!!), but in theory we're going to get two KEYDOWN events for most physical key presses now, right?

This indeed is introduced the problem that I now get double key events, e.g. in uhexen2's console.

sulix:
Yeah: that seems like it needs the events from TEXTINPUT and KEYDOWN sources to be merged.

I've tried hacking something together [here](https://github.com/sulix/sdl12-compat/tree/textinput-2):
https://github.com/sulix/sdl12-compat/commit/0d1842e4fda1ea6a8eb5cf5e86869d70e8f5cb92

It seems to work on uhexen2 for me, but I haven't tested it carefully on much else yet.

sezero:
> Yeah: that seems like it needs the events from TEXTINPUT and KEYDOWN sources to be merged.
> 
> I've tried hacking something together [here](https://github.com/sulix/sdl12-compat/tree/textinput-2):
> [sulix@0d1842e](https://github.com/sulix/sdl12-compat/commit/0d1842e4fda1ea6a8eb5cf5e86869d70e8f5cb92)
> 
> It seems to work on uhexen2 for me, but I haven't tested it carefully on much else yet.

Seems to work for me with uhexen2, quakespasm, and hhexen.

@icculus, @slouken?

icculus:
Looks good to me. Let's merge it.

icculus:
Oh, one note: if `EnabledUnicode` is zero, we should just ignore SDL_TEXTINPUT events and not hold the keydown events, as the `unicode` field is always expected to be zero in that case.

sulix:
Okay, I've made this check that `EnableUnicode` is set, and always flush the pending KEYDOWN event if it's not set.

I also tweaked it slightly to always put the keysym in the first simulated keydown event (it previously would put it in the second one if there were surrogate pairs).

I haven't explicitly made it ignore SDL_TEXTINPUT events — I assumed (perhaps wrongly) that with SDL_StartTextInput() not called, we shouldn't get any? I'll change this if that's not the case, though.

In any case, this is pull request #44 now.

Issue #17, opened 2021-02-11
Lack of unicode support breaks keyboard input in some applications.


SDLBugzilla:
# This bug report was migrated from our old Bugzilla tracker.

**Reported in version:** unspecified
**Reported for operating system, platform:** Linux, x86_64

# Comments on the original bug report:

On 2019-11-24 01:51:19 +0000, Lyle wrote:

> Some applications, such as the game Frozen Synapse and this legacy SDL tutorial from Lazy Foo http://www.lazyfoo.net/SDL_tutorials/lesson23/index.php pull keyboard input through event.key.keysym.unicode. Currently, this is always set to 0, which results in these applications failing to recognize any input. This occurs due to the following line in SDL12_compat.c:
> 
>     event12.key.keysym.unicode = 0;  FIXME("unicode");
> 
> However, since the unicode character set is a superset of ASCII, we can provide partial functionality here very easily just by passing through the sym value. For example:
> 
>     event12.key.keysym.unicode = event12.key.keysym.sym;
> 
> I'll admit that I don't know exactly what will happen here if someone does input a character which requires unicode support. However, I have verified that this change restores keyboard input functionality for me for the sign-in screen of the game Frozen Synapse as well as the above linked legacy Lazy Foo tutorial. Until such time that someone is prepared to fully implement the needed bits for unicode, I think this or something very much like it would be a change for the better.

sezero:
Reproduced the issue in quakespasm (http://quakespasm.sf.net/): 
typing in the game console doesn't work at all, because its SDL-1.2
code rely on event.key.keysym.unicode.

With the proposed change above, basic typing becomes possible
(but uppercase typing does not..)

sezero:
Implemented the proposed workaround as commit 65264db31389 for now.

sezero:
Should be fixed after commits 65264db31389c, b635c63ea4c01, ebfcc6e76383d,
and 35fa90b5d1a73. Closing.

PR #44, opened 2021-04-14
Attempt to merge KEYDOWN and TEXTINPUT events to prevent duplicates

sulix:
This is a follow-up to #41 and, together with it, should constitute a fix for #17 — keyboard support is working in most things with it, and hopefully other text input bugs will just need tweaks to this.

This change makes all KEYDOWN events be stored temporarily as a "pending keydown event". If a TEXTINPUT event comes in before the next time PumpEvents() is called (or another KEYDOWN or KEYUP event appears), then one of the events generated from the 'text' string is merged with the KEYDOWN event, thus having both a valid 'sym' and 'unicode' value.

Note that, if SDL_EnableUNICODE() is not enabled, these KEYDOWN events are all flushed immediately, which should be equivalent to the old behaviour.

Also add some basic support for UTF-16 surrogate pairs, as that was annoying me. I haven't actually found an SDL1.2 game with emoji support to test it with, but in theory, it should now work :-).

Note that this doesn't quite match what SDL 1.2 does, which depends pretty heavily on what backend was being used. The closest backend (XIM, I think) does basically the same thing as this, but puts the real keysym on the last character of a mult-character TEXTINPUT, whereas we put it in the first one. (The last one may actually be more sensible, in case
there are bugs where a 'last key pressed' is being maintained by an application, and the correct one is being overwritten by SDLK_UNKNOWN, but this was easier to implement for now.)

### DIFF
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -807,6 +807,8 @@ static EventQueueType *EventQueueHead = NULL;
 static EventQueueType *EventQueueTail = NULL;
 static EventQueueType *EventQueueAvailable = NULL;
 
+/* This is a KEYDOWN event which is being held for a follow-up TEXTINPUT */
+static SDL12_Event PendingKeydownEvent;
 
 /* Obviously we can't use SDL_LoadObject() to load SDL2.  :)  */
 static char loaderror[256];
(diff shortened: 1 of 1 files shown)


sezero:
Thanks.

PR #51, opened 2021-05-15
Support installing the library

Conan-Kudo:
This adds minimal logic to support installing the library correctly
on various platforms when the install target is being used.

This makes it possible to build and install it for both native and
cross compile targets for Linux, macOS, and Windows.


Conan-Kudo:
@icculus I'm not sure if you want this yet or if there's something else you're waiting on before supporting this...

sezero:
This is possibly the reason for https://github.com/libsdl-org/sdl12-compat/issues/85

PR #61, opened 2021-05-20
Support Multisampling with OpenGL Logical Scaling

sulix:
OpenGL Logical Scaling didn't work when Multisampling was enabled, as the faux-backbuffer didn't have multiple samples, so the format didn't match, making the blit fail and resulting in a black screen.

This change intercepts calls to set up MSAA in ``SDL_GL_SetAttribute()``, and, if OpenGL Logical Scaling is requested, creates an FBO with the desired number of samples, and a second FBO used for multisample resolve before scaling. (It's not possible for glBlitFramebuffer() to scale and resolve at the same time -- at least without the GL_EXT_framebuffer_multisample_blit_scaled extension.)

If OpenGL Logical Scaling has not been requested, we quickly SDL_GL_SetAttribute() the appropriate values for real before creating the window. This implementation does have the bug that MSAA won't work if OpenGL Logical Scaling is requested, but fails to initialise, as we've already created the window without it by the time we know we need to fall back.

Note that this patch doesn't really bother to handle GL_MULTISAMPLEBUFFERS separately from GL_MULTISAMPLESAMPLES — if the latter is > 0, the former is considered to be enabled, and vice-versa. This hasn't caused any problems thus far, but might be worth fixing…

It's also worth noting that this is using ``EXT_framebuffer_multisample``, and doesn't support the OpenGL 3 version of the functions without the EXT suffix — I'm yet to encounter an OpenGL 3+ system which doesn't support both, though, so it didn't seem like a big deal.

I've tested this with Psychonauts, and it works there. A few other OpenGL games didn't obviously break, too, but it probably could use more testing overall.


sulix:
I've made a couple of updates to this PR.

In particular, I'd got ``EXT_framebuffer_object`` and ``ARB_framebuffer_object`` mixed up: the latter has the multisample functions built-in (without the EXT suffix), so I've updated it to use those. I also now just pass 0 as the number of samples to ``glRenderbufferStorageMultisample()``, rather than checking and falling back to ``glRenderbufferStorage()``, which makes the code a bit simpler.

Unfortunately, I've also found a regression: [Cogs](https://www.cogsgame.com/) used to fallback to non-MSAA rendering with sdl12-compat when MSAA was enabled in its options, now it just shows a black screen. This is because it relies on ``glCopyTexImage2D()``, which doesn't work with the faux backbuffer. Basically, it only can read from multisampled framebuffers if they're the default framebuffer, not an FBO. Despite trying to work around this by using the previous frame's data, it actually does some rendering it requires immediately before calling ``glCopyTexImage2D()``, so the image in question hasn't been resolved anywhere.

The fix to this is to also have a shim for ``glCopyTexImage2D()`` which does a full multisample resolve before it's copied. I've got a hacky implementation for it in another branch here: https://github.com/sulix/sdl12-compat/commit/a2f2dede2ddc687e91bccc16e0e3a0d9bddb1af1
Of course, there are several other functions (``glReadPixels``, ``glCopyPixels``, etc) which would need shimming to handle all cases. The code also could do with a bit more optimisation to cache and check the currently bound framebuffers before triggering a resolve.

Personally, I still think this PR probably fixes more games than it breaks so far, but I can look into tidying up and adding more shims if that seems the way to go…

sezero:
@icculus: This is yours.

sulix:
I've rebased this on the latest ``main`` branch, and implemented shims for ``glReadPixels()``, ``glCopyPixels()``, ``glCopyTexImage1D()``, ``glCopyTexSubImage1D()``, ``glCopyTexImage2D()``, ``glCopyTexSubImage2D()``, and ``glCopyTexSubImage3D()`` which do an MSAA resolve if the default read framebuffer is bound.

The current bound framebuffers are now also cached inside sdl12-compat in the ``OpenGLCurrentReadFBO`` and ``OpenGLCurrentDrawFBO`` variables, which means we don't need to repeatedly call ``glGetInteger()`` everywhere to preserve the bound FBO state after resolves, and can easily restrict resolves to cases where the default read framebuffer is bound.

I haven't found anything that this doesn't work for (save for games which don't use ``SDL_GL_GetProcAddress()``, and so don't get the shims, and Logical Scaling is broken for them anyway). There may be some other functions which need shims — the various ``EXT_direct_state_access`` version of things come to mind ­— but I'm not aware of any SDL 1.2 apps using them.

PR #65, opened 2021-05-25
Use scaled mouse coordinates from SDL_RenderSetLogicalSize()

sulix:
While sdl12-compat scales the graphics for fullscreen windows running at a non-native resolution, mouse input is not scaled to match. Most games handle this relatively well — maybe the mouse gets "lost" off the right or bottom corners of the screen, which is only really obvious on dual-screen setups — but some will end up with the system/hardware cursor offset from where the game thinks it is.

For non-OpenGL applications, ``SDL_RenderSetLogicalSize()`` should scale the mouse coordinates for the window's events for us, but this has two problems:
1. We use ``SDL_SetEventFilter()`` to receive events, which runs before the Event Watch callback used by SDL_Renderer to scale the input.
2. Only events for the particular window are affected: the global coordinates returned by ``SDL_GetMouseState()`` aren't.

This PR switches to using ``SDL_AddEventWatch()`` (and updates it after ``SDL_RenderSetLogicalSize()`` to ensure it's always last on the list of watchers), and makes ``SDL_GetMouseState()`` return cached coordinates from event processing.

For OpenGL applications, none of this works because ``SDL_RenderSetLogicalSize()`` isn't being used, so maybe the correct way to do this is to scale mouse coordinates manually. This PR doesn't attempt to implement that, though. If we did wish to manually scale coordinates, we could either do so for both non-OpenGL windows and OpenGL windows (discarding these changes), or just do so for OpenGL windows. Both options seem reasonably sensible to me, so it's a matter of taste.

I've tested this with [Long Live The Queen](https://www.hanakogames.com/llq.shtml), which uses an SDL1.2/pygame-based RenPy engine — I expect other RenPy games of a similar age will behave similarly. Most of the other games I usually test are OpenGL-based, so aren't fixed by this.

Does this seem like a sensible approach?

### DIFF
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -1477,7 +1477,7 @@ Init12Video(void)
     SDL20_memset(EventStates, SDL_ENABLE, sizeof (EventStates)); /* on by default */
     EventStates[SDL12_SYSWMEVENT] = SDL_IGNORE;  /* off by default. */
 
-    SDL20_SetEventFilter(EventFilter20to12, NULL);
+    SDL20_AddEventWatch(EventFilter20to12, NULL);
 
     VideoDisplayIndex = GetVideoDisplay();
     SwapInterval = 0;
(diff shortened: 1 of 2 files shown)


sezero:
@icculus: This is yours.

src/SDL12_compat.c
```diff
@@ -4097,7 +4103,7 @@ UpdateRelativeMouseMode(void)
             if (MouseInputIsRelative) {
                 /* reset position, we'll have to track it ourselves in SDL_MOUSEMOTION events, since 1.2
                  *  would give you window coordinates, even in relative mode. */
-                SDL_GetMouseState(&MousePositionWhenRelative.x, &MousePositionWhenRelative.y);
+                SDL_GetMouseState(&MousePosition.x, &MousePosition.y);
```
icculus, line 4106: This is my bug, not yours, but I think this was supposed to be SDL20_GetMouseState(). Calling the 1.2 shim here just sets MousePosition to itself.

icculus (approved):
I think this is good.

src/SDL12_compat.c
sezero: Should we change this to `SDL20_GetMouseState()`?

icculus: oh, yes, I thought we fixed that. Sorry! Done now: 19823493da98acb0e938468daacd40bfdb974b0c

PR #69, opened 2021-05-30
Scale mouse events with OpenGL Logical Scaling

sulix:
For non-OpenGL windows, mouse coordinates are scaled by ``SDL_RenderSetLogicalSize()`, but with OpenGL Logical Scaling, we need to do this ourselves.

Extract the viewport calculations for OpenGL Logical Scaling into their own function, which is called on mouse events and used to adjust the mouse coordinates. Because they're adjusted early enough in the event processing loop, they're already scaled by the time MousePosition is set.

Also, fix the ``SDL_GetMouseState()`` call to be an ``SDL20_GetMouseState()`` call, and scale its coordinates.

Note that the scaled events are passed down to any other SDL2 event handlers that might exist, as we modify the SDL2 events in the filter. Since this basically matches what ``SDL_RenderSetLogicalSize()`` does, it hopefully won't be a problem.

This seems to work okay on the few games I tested it with: the cursor doesn't get lost in a great nothingness off the edge of the screen. It seems to also handle the case where the "logical" resolution is actually higher than the "physical" resolution.

Note that some changes would be required if we were to use ``SDL_ALLOW_HIGHDPI``, as the viewport code is the same for rendering and input events.

### DIFF
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -1902,6 +1902,69 @@ SDL_EventState(Uint8 type, int state)
     return retval;
 }
 
+static SDL_Rect
+GetOpenGLLogicalScalingViewport()
+{
+    float want_aspect, real_aspect;
+    int drawablew, drawableh;
+    SDL_Rect dstrect;
+    SDL20_GL_GetDrawableSize(VideoWindow20, &drawablew, &drawableh);
+
+    want_aspect = ((float) OpenGLLogicalScalingWidth) / ((float) OpenGLLogicalScalingHeight);
+    real_aspect = ((float) drawablew) / ((float) drawableh);
+
+    if (SDL20_fabsf(want_aspect-real_aspect) < 0.0001f) {
+        /* The aspect ratios are the same, just scale appropriately */
+        dstrect.x = 0;
+        dstrect.y = 0;
+        dstrect.w = drawablew;
+        dstrect.h = drawableh;
+    } else if (want_aspect > real_aspect) {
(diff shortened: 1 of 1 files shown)


sezero:
Is this still needed? (It doesn't apply cleanly anymore, either.)

icculus:
We probably still need this, but it's worth noting we just added an environment variable so end users can cram SDL_WINDOW_ALLOW_HIGHDPI in there.

sulix:
Yes, this change is definitely still required: without it, the mouse is unusable for OpenGL games which are using logical scaling and the hardware mouse cursor.

Note that this still isn't "perfect" in terms of usability: if the hardware mouse cursor isn't shown, the cursor does still apear to get "stuck" a bit at the edges when the screen is pillarboxed/letterboxed, due to the mouse cursor moving in the dead space at either side of the screens. This matches what ``SDL_RenderSetLogicalSize()`` does, and avoiding it would require having different paths for the hardware mouse cursor being visible and not, which I think gets too ugly to be worthwhile.

In any case, I've rebased this change on top of ``main``, and fixed up support for ``SDL_WINDOW_ALLOW_HIGHDPI``, and it seems to work.  [Uplink](http://uplink.co.uk/) is broken at fullscreen non-native resolutions without this PR, but works fine with it. (The High-DPI stuff is a little trickier to test than I'd like, as Wayland only exposes the one resolution on my machine, so many games — including uplink — refuse to set anything other than the native resolution, which doesn't need the Logical Scaling support.)

PR #80, opened 2021-06-02
Windows cleanups...

icculus:
This is a draft PR so I can watch Continuous Integration chew on this while I push new commits.


icculus:
ok, we're solid now.

PR #81, opened 2021-06-03
Add some common low-resolution modes to the modes list

sulix:
Some modern platforms (e.g. Wayland on KDE) don't expose any video modes other than the current desktop resolution. For games which expect a number of common VGA video modes to be available, this could result in there either being no way to change resolution, or the potentially game failing to start due to a mode not being available.

This change, as-is, currently adds three "fake" modes:
- 640×480
- 800×600
- 1024×768

These are added to the mode list for all available pixel formats, and
the SDL_VideoModeOK() function now uses the fake 1.2 modelist, so picks
up these modes.

Note that, thus far, these are always added to the beginning of the list of modes, whereas at least X11/XWayland on my machine return modes in largest→smallest mode. Maybe we want to add these modes at the end, then.

Also, maybe there are other common modes we should support (320×{200,240}, 1280×1024, maybe 1280×720 and 1920×1080). Do we want to only add these modes if the deskop resolution is higher? Or there exists an available mode with a higher resolution? Do we want to hide these behind an environment variable?

### DIFF
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -1439,6 +1439,58 @@ GetVideoDisplay(void)
     }
 }
 
+/* returns true if mode1 should sort before mode2 */
+static int
+VidModeSizeGreater(SDL12_Rect *mode1, SDL12_Rect *mode2)
+{
+    if (mode1->w > mode2->w) {
+        return 1;
+    } else if (mode2->w > mode1->w) {
+        return 0;
+    } else {
+        return (mode1->h > mode2->h);
+    }
+}
+
+static int
+AddVidModeToList(VideoModeList *vmode, SDL12_Rect *mode)
+{
+    void *ptr = NULL;
+    int i;
+
+    /* make sure we don't have this one already (with a different refresh rate, etc). */
+    for (i = 0; i < vmode->nummodes; i++) {
+        if ((vmode->modeslist12[i].w == mode->w) && (vmode->modeslist12[i].h == mode->h)) {
+            break;
+        }
+    }
+
+    if (i < vmode->nummodes) {
+        return 0;  /* already have this one. */
+    }
+
+    ptr = SDL20_realloc(vmode->modeslist12, sizeof (SDL12_Rect) * (vmode->nummodes + 1));
+    if (ptr == NULL) {
+        return SDL20_OutOfMemory();
+    }
+    vmode->modeslist12 = (SDL12_Rect *) ptr;
+
+    vmode->modeslist12[vmode->nummodes] = *mode;
+
+    vmode->nummodes++;
+
+    return 0;
+}
+
+/* A list of fake video modes which are included. */
+static SDL12_Rect fake_modes[] = {
+    { 0, 0, 1920, 1080 },
+    { 0, 0, 1280, 720 },
+    { 0, 0, 1024, 768 },
+    { 0, 0, 800, 600 },
+    { 0, 0, 640, 480 }
+};
+
 /* This sets up VideoModes and VideoModesCount. You end up with arrays by pixel
     format, each with a value that 1.2's SDL_ListModes() can return. */
 static int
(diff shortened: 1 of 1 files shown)


slouken:
If I remember right, the current desktop mode was first, and the other modes were listed largest to smallest.

I think the logic should add the fake modes in that order, and of course skip any fake modes that are larger than the current desktop resolution.

icculus:
The 1.2 docs say the list is always sorted largest to smallest (the X11 driver sorts it largest to smallest width, and if two modes have the same width, largest to smallest height).

SDL2 gives this order, dropping to the next item on the list when something is equal.
- width -> largest to smallest
- height -> largest to smallest
- bits per pixel -> more colors to fewer colors
- packed pixel layout -> largest to smallest
- refresh rate -> highest to lowest

I don't think either promise the desktop size first, but the docs could be wrong (or I've just always had my display at the highest resolution possible, which is likely).

Also, fake modes without logical GL scaling is going to fail, I think? With software surfaces we always scale, so no problem there...I guess if someone forces logical scaling off and gets bit by that problem, it's their own fault, though.

sulix:
I've uploaded a new version which sorts the modes according to width, then height. Modes are included if a higher-resolution mode also exists (even if the current desktop resolution is smaller). The implementation is a little bit uglier than I'd like it to be at the moment, with the list of modes being included twice, though.

A super-quick check on my machine showed the modes being sorted according to size in SDL 1.2, even when the desktop resolution is not the largest mode.

We could disable the fake modes if ``SDL12COMPAT_OPENGL_SCALING`` is explicitly disabled by the user, I guess — it does tend to fail in some way: everything I tried just ended up rendering into the bottom-left corner of the screen.

I also added 1920×1080 and 1280×720, as they're sufficiently standard resolutions, and this makes using wayland on my laptop much more natural (and, interesting, made old SDL 1.2 builds of Braid work correct for the first time). To compare the before-and-after outputs of ``SDL_VIDEODRIVER=wayland ./testvidinfo``:

Before:
```
Video driver: wayland
Current display: 3200x1800, 32 bits-per-pixel
        Red Mask = 0x00ff0000
        Green Mask = 0x0000ff00
        Blue Mask = 0x000000ff
Fullscreen video modes:
        3200x1800x32
A window manager is available
```

After:
```
Video driver: wayland
Current display: 3200x1800, 32 bits-per-pixel
        Red Mask = 0x00ff0000
        Green Mask = 0x0000ff00
        Blue Mask = 0x000000ff
Fullscreen video modes:
        3200x1800x32
        1920x1080x32
        1280x720x32
        1024x768x32
        800x600x32
        640x480x32
A window manager is available
```

Interestingly, XWayland seems to give a large number of modes (probably faked) of its own on this machine (due to 200% DPI scaling, all the sizes are halved from wayland):
```
Video driver: x11
Current display: 1600x900, 32 bits-per-pixel
        Red Mask = 0x00ff0000
        Green Mask = 0x0000ff00
        Blue Mask = 0x000000ff
Fullscreen video modes:
        1600x900x32
        1440x900x32
        1368x768x32
        1280x800x32
        1280x720x32
        1152x864x32
        1024x768x32
        1024x576x32
        864x486x32
        800x600x32
        720x480x32
        720x400x32
        640x480x32
        640x400x32
        640x350x32
        320x240x32
        320x200x32
A window manager is available
```

sulix:
I've tidied up some of how this works: there's now a ``fake_modes`` array of ``SDL12_Rect`` which contains the list of fake modes to insert. The helper functions now all take ``SDL12_Rect`` directly, which looks a bit cleaner.

It also now disables the fake modes if ``SDL12COMPAT_OPENGL_SCALING=0`` is set. They'll still appear if we fallback to a non-scaled path if the environment variable isn't set, though, as we're initialising the mode list before the window is created.

In any case, I think this is probably ready: it seems to work fine here. The ``testvidinfo`` results are the same as in my previous comment: the actual behaviour shouldn't have changed at all since the previous revision (unless logical scaling is disabled).

Conan-Kudo:
@sulix Is this still WIP though? You have it marked as such in the title...

sulix:
> @sulix Is this still WIP though? You have it marked as such in the title...

Oops… I forgot to change that. Assuming everyone's okay with it, I think this is good to go.

PR #82, opened 2021-06-03
Fix sdl-config and sdl12_compat contents

lanodan:


sezero:
Our CMake'ry to additionally create a `libSDL.so` (or `libSDL.dylib`) symlink
like the standalone makefiles do (see e.g. https://github.com/libsdl-org/sdl12-compat/commit/f6a60d6534a16308cfc22f7b0945797849efb083)

Conan-Kudo (approved):

lanodan:
Ended up rewriting the bits for `sdl-config` and `sdl12-compat.pc` a bit more so they should now have exactly the same content.

Symlink for `-lSDL` to work got added.

Conan-Kudo (approved):

CMakeLists.txt
```diff
@@ -160,32 +160,36 @@ install(TARGETS SDL
...
-  set(exec_prefix "\${prefix}")
-  set(libdir "\${exec_prefix}/lib${LIB_SUFFIX}")
-  set(bindir "\${exec_prefix}/bin")
-  set(includedir "\${prefix}/include")
   configure_file("${CMAKE_SOURCE_DIR}/sdl-config.in" "${CMAKE_BINARY_DIR}/sdl-config" @ONLY)
   install(PROGRAMS "${CMAKE_BINARY_DIR}/sdl-config" DESTINATION bin)
+
+  set(SOEXT ${CMAKE_SHARED_LIBRARY_SUFFIX})
+  get_target_property(SONAME SDL OUTPUT_NAME)
+  if(NOT ("${SONAME}" EQUAL "SDL"))
```
sezero, line 187: I guess this needs guarding with something like `if(NOT (WIN32 OR CYGWIN))`

lanodan: Well it could be an `if(UNIX AND NOT ANDROID)` given the `OUTPUT_NAME` logic in CMakeLists.txt:54 but I think it's better to avoid duplicating that logic.

sezero: @icculus, this is yours

icculus: I guess wrap it in (UNIX AND NOT ANDROID) for now.

icculus (changes requested):

PR #93, opened 2021-06-09
cmake: Add support to build the static library

Conan-Kudo:
A static link library is not ideal, since we cannot statically link
SDL2 into this library. However, some projects demand a static library
to build, so optionally provide a way to build it so those projects
can use this library instead of SDL-1.2.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,7 @@ set(SDL12_COMPAT_VERSION_STR "1.2.50")
 
 option(SDL12TESTS "Enable to build SDL-1.2 test programs" ON)
 option(SDL12DEVEL "Enable installing SDL-1.2 development headers" ON)
+option(STATICDEVEL "Enable installing static link library" OFF)
 
 set(CMAKE_SKIP_RPATH TRUE)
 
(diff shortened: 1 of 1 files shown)


sezero:
Link libraries, linker flags and mac rpath shouldn't be necessary.

Those aside, I don't like this one - allowing a static library...

Conan-Kudo:
If we were generating CMakeConfig target things, it would probably be more meaningful, I suppose? That said, I left the static library off by default on purpose...

sezero:
> If we were generating CMakeConfig target things, it would probably be more meaningful, I suppose?

You mean the link libraries? Maybe, I am not a cmake guy after all.
But certainly not the link flags, e.g. `/NODEFAULTLIB` or `-nostdlib`, yes?

> That said, I left the static library off by default on purpose...

@icculus to decide.

icculus:
> @icculus to decide.

I believe the intention here was to aid Fedora's migration, but looking at the existing RPMs, they don't appear to have .a files included. Do we _actually_ need this, @Conan-Kudo?

(If they need it, I'm fine with it being included and disabled by default...but practically speaking, it's not hard to statically link a single C file into a project, so there doesn't seem to be a lot of value in trying to _prevent_ people from doing it by not including it in the CMake file.)

Conan-Kudo:
@icculus It was brought up by the SDL maintainer that we need to have it available to fully replace that package, since the old SDL package provides the static library.

icculus:
Oh, I didn't realize it was a separate `static` package, I was looking at the wrong RPM.

Okay, this goes in, then, disabled by default.

icculus:
> Okay, this goes in, then, disabled by default.

(and if this breaks on Windows or whatever, the person that enabled it can send a bug report or a patch.)

Issue #84, opened 2021-06-06
dosbox keyboard layout issues


sezero:
dosbox-0.74 on CentOS-6.10 / i686.

With 'normal' SDL-1.2, dosbox doesn't respect my keyboard layout (which
is Turkish-F) and behave as is I were using US layout.

With sdl12-compat (latest git master) talking to SDL2-latest git master,
it does use my keyboard layout but not fully correctly:

- The key next to left-shift should report `<`, `>` or `|` (plain, with
  shift, with AltGr) but it prints `\`, `|` or none: all behaving like
  US-layout.

- The key next to right shift should report `,` or `;` (plain or shift),
  but it prints `,` (which is good) or `<` (bad: US-layout).

- The key left to it should report `.` or `:`, but it prints `.` (good)
  or `>` (bad: US-layout).

- Similar stuff with the keys on the numbers' row -- e.g. the key below
  Esc prints nothing, ...

sezero:
If it matters, here is the console output:
```
INFO: sdl12-compat, built on Jun  6 2021 at 19:44:55, talking to SDL2 2.0.15
DOSBox version 0.74
Copyright 2002-2010 DOSBox Team, published under GNU GPL.
---
INFO: FIXME: there is never a parachute in SDL2, should we catch segfaults ourselves? (SDL_Init:1578)
INFO: FIXME: support SDL_INIT_EVENTTHREAD where it makes sense? (SDL_InitSubSystem:1552)
CONFIG:Loading primary settings from config file /home/ozzie/.dosbox/dosbox-0.74.conf
INFO: FIXME: Should we offer scaling for windowed modes, too? (SDL_SetVideoMode:3594)
INFO: FIXME: currently ignores SDL_WINDOWID, which we could use with SDL_CreateWindowFrom ...? (SDL_SetVideoMode:3611)
INFO: FIXME: handle SDL_ANYFORMAT (SDL_SetVideoMode:3625)
INFO: FIXME: There's an environment variable to choose a display (SDL_SetVideoMode:3632)
INFO: FIXME: bgr instead of rgb? (SDL_SetVideoMode:3652)
INFO: FIXME: don't do anything if the window's dimensions, etc haven't changed. (SDL_SetVideoMode:3659)
INFO: FIXME: we need to preserve VideoSurface12 (but not its pixels), I think... (SDL_SetVideoMode:3660)
INFO: FIXME: setup screen saver (SDL_SetVideoMode:3899)
[...]

PR #97, opened 2021-06-10
Convert SDL2 Scancodes into SDLKey, not SDL2 Keycodes

sulix:
[*WARNING*: This completely changes a lot of the keyboard input behaviour, and so is quite likely to cause some regressions. It seems to work okay for me, but there may be implications of this I've not noticed…]

SDL 1.2's keysym appears to be entirely layout-indpendent: it always acts as though there is a US keyboard layout. i.e., the values represent the positions on the keyboard, not the "meaning" of the key. This matches what SDL 2.0 scancodes do, but not SDL 2.0 keycodes, which are translated according to the current keyboard layout.

This means that many SDL 1.2 applications would suddenly start using the user's keyboard layout, rather than the US one, which -- while quite convenient in some cases -- is a change in behaviour and could break things.

Instead of converting SDL_Keycode→SDLKey, this patch converts from SDL_Scancode→SDLKey, preserving the original behaviour. SDL 1.2 applications will still get translated values in the 'unicode' Keysym variable if enabled, so text input in many applications will still work as expected.

On the implementation side, this currently just has the one ``CASESCANCODE20TOKEY12`` function which takes both the SDL_Scancode and SDLKey names, even though these are identical most of the time. The previous implementation had a separate macro for the case there were the same: I found it easier to copy-paste things without it, but I'm happy to do that if we prefer.

The other option is to keep both implementations, and have an environment variable to switch between them. That might allow us to add keyboard layout support to some SDL 1.2 titles which previously didn't support it, but could break some things (e.g. DOSBox's own keyboard layout support with ``usescancodes=false``, as seen in #84) if we used Keycodes by default.


icculus:
Ugh, I think this is probably the right thing to do, but also it's a little gross to be nerfing the functionality to make it more correct.

@slouken and @sezero, do you have an opinion on this one?

slouken:
Yes, this is probably correct, I'll go ahead and merge it.

sezero:
Since this patch has been merged (as far as I can tell), [hhexen](https://github.com/sezero/hhexen.git) doesn't
respect my keyboard layout and behaves as if I were using a US-layout.

Its SDL keyboard handling code is here: https://github.com/sezero/hhexen/blob/master/sdl/i_sdlgl.c#L217

sezero:
Same with uhexen2, or quakespasm. And the issue is with the in-game key hits: text input seems fine.

Was that really the intention?

sezero:
Pushed https://github.com/libsdl-org/sdl12-compat/commit/0bde9879e87f3c6835e7191ae43357900f3f3641 to help testing: It is still the new behavior.
If you change the `#if 0` to `#if 1` around line 2432, it is the old behavior.

Comments about what to do?

sulix:
> Since this patch has been merged (as far as I can tell), [hhexen](https://github.com/sezero/hhexen.git) doesn't
> respect my keyboard layout and behaves as if I were using a US-layout.

Yes, this was intentional — I think this is what the original SDL 1.2 will do, as well. (I tested with QuakeSpasm, and that seemed to be the case.) It's a shame, because I definitely prefer the old behaviour where it works.

I'm not opposed to making this an option, but since DOSBox (and presumably some other things) require it, it's probably worth having this behaviour as the default.

I guess the other option, for games which make use of it, would be to make ``SDL_GetKeyName()`` do the keyboard layout translation. I'm not sure if the real SDL 1.2 does something like that?

sezero:
Real SDL-1.2 does respect the kbd layout

sulix:
Huh! It doesn't on my machine (except for the ``unicode`` value)! 

The console in Quakespasm does (because it uses the ``event.key.keysym.unicode`` value), but the actual in-game keys don't (indeed, there's a comment explaining that it's deliberate that it doesn't, at least for SDL2: https://github.com/sezero/quakespasm/blob/master/Quake/in_sdl.c#L1052 )

Does ``checkkeys`` give different results when run under SDL 1.2 and sdl12-compat? It gives the exact same output for me, regardless of keyboard layout.

sezero:
checkkeys built from SDL-1.2-git, with my Turkish-F layout,

run against SDL-1.2.14 (as provided by CentOS-6.10):
```
Key pressed:  102-f (scancode = 24 [0x18]) (f) modifiers: NUM
Key released:  102-f (scancode = 24 [0x18]) modifiers: NUM
Key pressed:  103-g (scancode = 25 [0x19]) (g) modifiers: NUM
Key released:  103-g (scancode = 25 [0x19]) modifiers: NUM
Key pressed:  27-escape (scancode = 9 [0x9]) (^[) modifiers: NUM
Key released:  27-escape (scancode = 9 [0x9]) modifiers: NUM
```

run against SDL-1.2.16 (SDL-1.2-git):
```
Key pressed:  102-f (scancode = 24 [0x18]) (f) modifiers: NUM
Key released:  102-f (scancode = 24 [0x18]) modifiers: NUM
Key pressed:  103-g (scancode = 25 [0x19]) (g) modifiers: NUM
Key released:  103-g (scancode = 25 [0x19]) modifiers: NUM
Key pressed:  27-escape (scancode = 9 [0x9]) (^[) modifiers: NUM
Key released:  27-escape (scancode = 9 [0x9]) modifiers: NUM

```

run against sdl12-compat git:
```
Key pressed:  113-q (scancode = 24 [0x18]) (f) modifiers: NUM
Key released:  113-q (scancode = 24 [0x18]) modifiers: NUM
Key pressed:  119-w (scancode = 25 [0x19]) (g) modifiers: NUM
Key released:  119-w (scancode = 25 [0x19]) modifiers: NUM
Key pressed:  27-escape (scancode = 9 [0x9]) (^[) modifiers: NUM
Key released:  27-escape (scancode = 9 [0x9]) modifiers: NUM
```

sulix:
Huh — that's very strange. I get this:
```
Key pressed:  113-q (scancode = 24 [0x18]) (f) modifiers: (none)
Key released:  113-q (scancode = 24 [0x18]) modifiers: (none)
Key pressed:  119-w (scancode = 25 [0x19]) (g) modifiers: (none)
Key released:  119-w (scancode = 25 [0x19]) modifiers: (none)
Key pressed:  27-escape (scancode = 9 [0x9]) (^[) modifiers: (none)
Key released:  27-escape (scancode = 9 [0x9]) modifiers: (none)
```
Using the Turkish F layout on SDL 1.2.15 (Ubuntu 21.10/impish), SDL 1.2.16 (current git) and current sdl12-compat.

The only thing I can think could be causing issues is that I have the compose key enabled, which sometimes breaks complex keyboard layouts. This is also under regular X11 (with KDE) rather than wayland, so that's hopefully normal enough.

xev shows that, for US English, I get ``keycode 24 (keysym 0x71, q)`` for "q", but with Turkish F, I get ``keycode 24 (keysym 0x66, f)``, which shows that X11 is translating the keysym correctly, and leaving the keycode the same. Regardless, it definitely looks like something strange is going on with one or both of our systems, then…

sezero:
The above was with x11+gnome.

The following is with x11 (xwayland I _think_?) and gnome3 on fedora-27,
using Turkish Q layout:

against SDL-1.2.15 (as provided by fedora):
```
Key pressed:  187-world 27 (scancode = 34 [0x22]) modifiers: (none)
Key released:  187-world 27 (scancode = 34 [0x22]) modifiers: (none)
Key pressed:  252-world 92 (scancode = 35 [0x23]) (�) modifiers: (none)
Key released:  252-world 92 (scancode = 35 [0x23]) modifiers: (none)
Key pressed:  231-world 71 (scancode = 60 [0x3c]) (�) modifiers: (none)
Key released:  231-world 71 (scancode = 60 [0x3c]) modifiers: (none)
Key pressed:  46-. (scancode = 61 [0x3d]) (.) modifiers: (none)
Key released:  46-. (scancode = 61 [0x3d]) modifiers: (none)
Key pressed:  27-escape (scancode = 9 [0x9]) (^[) modifiers: (none)
Key released:  27-escape (scancode = 9 [0x9]) modifiers: (none)
```

against sdl12-compat from git:
```
Key pressed:  91-[ (scancode = 34 [0x22]) (0x11F) modifiers: NUM
Key released:  91-[ (scancode = 34 [0x22]) modifiers: NUM
Key pressed:  93-] (scancode = 35 [0x23]) (�) modifiers: NUM
Key released:  93-] (scancode = 35 [0x23]) modifiers: NUM
Key pressed:  46-. (scancode = 60 [0x3c]) (�) modifiers: NUM
Key released:  46-. (scancode = 60 [0x3c]) modifiers: NUM
Key pressed:  47-/ (scancode = 61 [0x3d]) (.) modifiers: NUM
Key released:  47-/ (scancode = 61 [0x3d]) modifiers: NUM
Key pressed:  27-escape (scancode = 9 [0x9]) (^[) modifiers: NUM
Key released:  27-escape (scancode = 9 [0x9]) modifiers: NUM
```

However, if I set it up to switch between F and Q layouts and select F,
the following happens with SDL-1.2.15:
```
Key pressed:  113-q (scancode = 24 [0x18]) (f) modifiers: (none)
Key released:  113-q (scancode = 24 [0x18]) modifiers: (none)
Key pressed:  119-w (scancode = 25 [0x19]) (g) modifiers: (none)
Key released:  119-w (scancode = 25 [0x19]) modifiers: (none)
Key pressed:  27-escape (scancode = 9 [0x9]) (^[) modifiers: (none)
Key released:  27-escape (scancode = 9 [0x9]) modifiers: (none)
```
So I'm not sure what's going on..

sezero:
Addendum to above: If I set it up not to switch layouts but only one
which is the F layout, then SDL-1.2.15 does respect my F layout.

sezero:
@slouken, @icculus ?

sulix:
I tested this on Windows 10 (though, admittedly in a VM on the Linux machine) as well, and SDL 1.2 it also does not take the keyboard layout into account:
```
Key pressed:  113-q (scancode = 16 [0x10]) (f) modifiers: (none)
Key released:  113-q (scancode = 16 [0x10]) modifiers: (none)
Key pressed:  119-w (scancode = 17 [0x11]) (g) modifiers: (none)
Key released:  119-w (scancode = 17 [0x11]) modifiers: (none)
Key pressed:  27-escape (scancode = 1 [0x1]) (^[) modifiers: (none)
Key released:  27-escape (scancode = 1 [0x1]) modifiers: (none)
```

This makes me think that the correct behaviour is for sdl12-compat to not take the keyboard layout into account, and that what you're seeing is a bug in SDL 1.2 which probably only shows up on X11 if there is only one, non-QWERTY layout available.

Maybe it's worth leaving a config option around if people turn out to be relying on this bug, but I don't think that sdl2-compat's default behaviour should change, since we know that at least DOSBox depends on the keysym being in the US layout.

DanielGibson:
I'm pretty sure SDL 1.2 took my (German QWERTZ) keyboard layout into account for keycodes, though you people with US layouts of course can't tell the difference :-P

For example, look at SDL1.2's `X11_TranslateKeycode()`: https://github.com/libsdl-org/SDL-1.2/blob/main/src/video/x11/SDL_x11events.c#L1155
It translates the X11 KeyCode (which is kinda like a SDL scancode) to a X11 KeySym (which is supposed to be the thing printed on the keyboard's keycaps, i.e. like SDL2 SDL_KeyCode, see https://tronche.com/gui/x/xlib/input/keyboard-encoding.html) and then translates that to a `SDLKey` which is used in `SDL_keysym::sym` (see https://github.com/libsdl-org/SDL-1.2/blob/main/src/video/x11/SDL_x11events.c#L606 ).

So `SDLKey` indeed is based on the name of (printed on) the key and not the scancode.

So please revert this pull request and map from SDL2 SDL_KeyCode to SDL12Key again.

sezero:
Does changing that `#if 0` to `#if 1` work as it should, as noted at https://github.com/libsdl-org/sdl12-compat/pull/97#issuecomment-867013649 ?

DanielGibson:
It should, at least for the keys handled there.

I hope I'll find some time to test and possibly improve this in the foreseeable future, but I can't promise anything right now.
(TBH I just saw a tweet mentioning Fedora replacing SDL1.2 with sdl12-compat and wondered if keyboard input is "solved" now and looked at the source to find out it isn't and saw this PR linked and was quite surprised that no one except for you seemed to remember that SDL1.2 indeed handled non-US keyboard layouts)

PR #98, opened 2021-06-10
Add the Right Ctrl, Alt, Shift and Meta key scancodes on macOS

sulix:
It turns out that, as of Mac OS X Panther, there is the ability to distinguish between left and right Ctrl, Alt, Shift, and Meta keys, and scancodes were assigned. DOSBox doesn't seem to be aware of this, so we didn't add them when setting up the scancodes in the first place.

See the patch adding them to SDL 1.2 here:
https://github.com/libsdl-org/SDL-1.2/commit/85d87922f4f1315230f55a12c09dd46bfba04f6d

Note that, of these, Right Ctrl and Alt are different on Linux evdev, so are mapped to 0 like other similar keys, as DOSBox otherwise will attempt to dynamically change the scancode it's looking for based on the X11 driver. Using 0 triggers a fallback to a non-scancode code path, which works fine.

See issue #84 for details.


sulix:
This looks to actually be wrong: SDL 1.2 never produces the correct scancodes for modifier keys on Mac OS X, instead always setting ``scancode`` to 0 (which, annoyingly, is the valid scancode for ``a``).

This can be seen in the SDL 1.2 source code:
https://github.com/libsdl-org/SDL-1.2/blob/main/src/video/quartz/SDL_QuartzEvents.m#L475

A more detailed analysis of what was happening with DOSBox is in the relevant issue:
https://github.com/libsdl-org/sdl12-compat/issues/84#issuecomment-858597299

PR #102, opened 2021-06-11
Close audio device in SDL_Quit{,Subsystem}()

sulix:
Given we allocate and use a bunch of wrappers around the SDL audio device, we should shut it down explicitly on quit.

The simplest way to implement this was to just call our ``SDL_CloseAudio()`` implementation in ``SDL_QuitSubsystem()``, which require making it robust against closing an already closed audio device.

This fixes a sporadic null pointer access in the audio callback on DOSBox on one of my machines.

### DIFF
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -1710,6 +1710,10 @@ SDL_QuitSubSystem(Uint32 sdl12flags)
         QuitCDSubsystem();
     }
 
+    if (sdl12flags & SDL12_INIT_AUDIO) {
+        SDL_CloseAudio();
+    }
+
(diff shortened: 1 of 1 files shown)


icculus:
I'm on a phone, so I can't check, but do we need to shutdown cd-rom the same way?

sulix:
It should be okay: there's already a call to ``QuitCDSubsystem()`` in ``SDL_QuitSubsystem()``.

I did some very basic testing to make sure that this patch at the very least didn't break the cd-rom. I don't know if we're actually stopping the cd playback if we quit the cd subsystem, though — I guess SDL 1.2 didn't on actual cd drives, as they'd happily keep playing even if the program quit…

PR #113, opened 2021-06-22
Don't set {R,G,B,A}masks on SDL2 paletted surfaces

sulix:
SDL2 expects 8-bit surfaces to have correct {R,G,B,A}masks in their format: either all 0 for paletted surfaces, or the correct values for ``SDL_PIXELFORMAT_RGB332``. Indeed, ``SDL_ConvertSurface()`` will crash if this
is not the case.

Some SDL 1.2 applications provide incorrect masks for paletted surfaces. We originally zeroed these out in 46fdbf7c1d, but 3af751b91b actually re-sets these in ``SetPalette12ForMasks()``.

Instead, do not set the masks on the SDL 2 surface in ``SetPalette12ForMasks()``, and force them to be 0 in ``PixelFormat12to20()` if there is a palette. The masks are still set on the SDL 1.2 surface's format, and the palette generated by SetPalette12ForMasks() should in theory make sure surfaces with 8-bit masks still work.

The other option, rather than making this depend on if the surface has a palette, would be to condition the zeroing of masks on the bit-depth, but I think this is better.

This fixes a crash on startup in circuslinux:
http://www.newbreedsoftware.com/circus-linux/

### DIFF
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -1415,18 +1415,35 @@ PixelFormat12to20(SDL_PixelFormat *format20, SDL_Palette *palette20, const SDL12
     format20->format = SDL20_MasksToPixelFormatEnum(format12->BitsPerPixel, format12->Rmask, format12->Gmask, format12->Bmask, format12->Amask);
     format20->BitsPerPixel = format12->BitsPerPixel;
     format20->BytesPerPixel = format12->BytesPerPixel;
-    format20->Rmask = format12->Rmask;
-    format20->Gmask = format12->Gmask;
-    format20->Bmask = format12->Bmask;
-    format20->Amask = format12->Amask;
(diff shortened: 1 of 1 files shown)


src/SDL12_compat.c
```diff
@@ -3258,16 +3275,6 @@ SetPalette12ForMasks(SDL12_Surface *surface12, const Uint32 Rmask, const Uint32
             color->a = 255;
         }
 
-        format20 = surface12->surface20->format;
```
sezero, line 3261: The `format20` local var can be removed now in this procedure

sulix: Removed, thanks!

sulix:
Pushed a new version with the ``format20`` variable removed in ``SetPalette12ForMasks()``. (Also rebased on top of ``main``.)

PR #123, opened 2021-07-25
Add `-lmingw32` to `SDL_LIBS`

Conan-Kudo:
This prevents undefined references to `WinMain@16` by consumers.

(From @manisandro for [the Fedora `mingw-sdl12-compat` package](https://src.fedoraproject.org/rpms/mingw-sdl12-compat/c/60d858f997a3f0ca137ff1dd2135c462cf0eda21?branch=rawhide))

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -234,7 +234,7 @@ if(SDL12DEVEL)
     if(WIN32)
       set(SDL_CFLAGS "")
       set(SDL_RLD_FLAGS "")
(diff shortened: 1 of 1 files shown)


slouken:
-lSDLmain is supposed to take care of that. Why isn't that working?

Conan-Kudo:
How would that work when `libSDLmain.a` is a static library? Static libraries typically can't resolve symbols on their own.

PR #124, opened 2021-07-26
Scale relative mouse coordinates when OpenGL Logical Scaling is in use

sulix:
The SDL_Render-based logical scaling (using ``SDL_RenderSetLogicalSize()``) will scale both the relative and absolute mouse inputs, whereas we were only scaling the absolute mouse position. To match the SDL_Render mode, this change adds relative mouse motion scaling to the OpenGL Logical Scaling code, as well as the ability to disable it with the ``SDL_MOUSE_RELATIVE_SCALING`` environment variable (already supported by SDL_Render).

Note that the actual difference in behaviour here is extremely subtle, and hasn't been causing any actual issues as far as I can tell, but it could "feel" a little bit weird, particularly on games which were using relative mouse motion to control a cursor.

Most of these issues were fixed in SDL 2.0 as a result of this bug report:
https://github.com/libsdl-org/SDL/issues/3407

I think this is good to go, but the interaction with the ``SDL_MOUSE_RELATIVE_SCALING`` hint could be changed to either:
- Use the SDL 2.0 hint system entirely, and use ``SDL20_GetHintBoolean()`` instead of ``getenv()``.
- Drop the use of the hint for the SDL_Render backend as well, and wrap it in a new environment variable with a new name (and set ``SDL_HINT_MOUSE_RELATIVE_SCALING`` based on that).

(But personally, I think I prefer it like this…)

### DIFF
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -166,6 +166,11 @@ SDL20_MostSignificantBitIndex32(Uint32 x)
 #endif
 }
 
+/* SDL_truncf needs SDL >=2.0.14 (and breaks Watcom), so copy it here */
+static float SDL20_truncf(float x)
+{
+    return (x < 0.0f) ? SDL20_ceilf(x) : SDL20_floorf(x);
+}
 
 #define SDL12_DEFAULT_REPEAT_DELAY 500
 #define SDL12_DEFAULT_REPEAT_INTERVAL 30
@@ -929,6 +934,7 @@ static JoystickOpenedItem JoystickOpenList[16];
 static Uint8 KeyState[SDLK12_LAST];
 static SDL_bool MouseInputIsRelative = SDL_FALSE;
 static SDL_Point MousePosition = { 0, 0 };
+static SDL_FPoint MouseRelativeRemainder = { 0.f, 0.f };
 static SDL_bool UseMouseRelativeScaling = SDL_FALSE;
 static OpenGLEntryPoints OpenGLFuncs;
 static int OpenGLBlitLockCount = 0;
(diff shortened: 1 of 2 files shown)


sulix:
(Oh — I should also point out that I haven't tested this on OpenWatcom — or indeed anything but Linux — so I have no idea if the hastily written fallback for ``SDL20_truncf`` works or compiles. So be warned!)

sezero:
The truncf stuff requires SDL >= 2.0.14, so a local replacement is more suitable for it.
Something llike the following, on top of your patch (which also handles watcom which
doesn't have trunc):
```diff
diff --git a/src/SDL12_compat.c b/src/SDL12_compat.c
index 72d3f6c..305d3ef 100644
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -2222,7 +2222,12 @@ AdjustOpenGLLogicalScalingPoint(int *x, int *y)
     *x = SDL_max(SDL_min(adjusted_x, OpenGLLogicalScalingWidth), 0);
     *y = SDL_max(SDL_min(adjusted_y, OpenGLLogicalScalingHeight), 0);
 }
-    
+
+static float SDL20_truncf(float x)
+{
+    return (x >= 0.0f) ? SDL20_floorf(x) : SDL20_ceilf(x);
+}
+
 /* Scale a vector (e.g. relative mouse movement) to the logical scaling size */
 static void
 AdjustOpenGLLogicalScalingVector(int *x, int *y, float *rx, float *ry)
diff --git a/src/SDL20_syms.h b/src/SDL20_syms.h
index 6540d5d..f1cb785 100644
--- a/src/SDL20_syms.h
+++ b/src/SDL20_syms.h
@@ -308,14 +308,14 @@ SDL20_SYM(int,atoi,(const char *a),(a),return)
 #ifndef SDL12_MATH
 #include <math.h>
 #define SDL20_fabsf fabs
+#define SDL20_ceilf ceil
 #define SDL20_floorf floor
-#define SDL20_truncf trunc
 #define SDL12_MATH
 #endif
 #else
 SDL20_SYM(float,fabsf,(float a),(a),return)
+SDL20_SYM(float,ceilf,(float a),(a),return)
 SDL20_SYM(float,floorf,(float a),(a),return)
-SDL20_SYM(float,truncf,(float a),(a),return)
 #endif
 
```

sulix:
Thanks — I've pushed a new version which includes a trunc() implementation.

sezero:
It's now good with SDL2 version requirements (and also watcom-wise.)
@icculus ?

Issue #127, opened 2021-08-12
Windowing issues and high CPU usage in Dwarf Fortress


autumnontape:
When starting Dwarf Fortress on X.Org or Xwayland with sdl12-compat, the game immediately hangs on a black screen, using ~95% of one CPU (with X.Org or Xwayland using the rest) and logging this to stdout:

<details>

```
Sound devices available:
OpenAL Soft
Picking OpenAL Soft. If your desired device was missing, make sure you have the appropriate 32-bit libraries installed. If you wanted a different device, configure ~/.openalrc appropriately.
Perfect OpenAL context attributes GET
Loading bindings from data/init/interface.txt
New window size: 640x300
Font size: 8x12
Resizing grid to 80x25
Resizing font to 8x12

Resetting textures
New window size: 640x300
Font size: 8x12
Resizing grid to 80x25
Resizing font to 8x12

New window size: 640x300
Font size: 8x12
Resizing grid to 80x25
Resizing font to 8x12

New window size: 640x300
Font size: 8x12
Resizing grid to 80x25
Resizing font to 8x12

New window size: 640x300
Font size: 8x12
Resizing grid to 80x25
Resizing font to 8x12

New window size: 640x300
Font size: 8x12
Resizing grid to 80x25
Resizing font to 8x12

New window size: 640x300
Font size: 8x12
Resizing grid to 80x25
Resizing font to 8x12

New window size: 640x300
Font size: 8x12
Resizing grid to 80x25
Resizing font to 8x12
```

\[repetition continues...]
</details>

It doesn't respond to the X button or normal kill signals and must be killed with `kill -9` when in this state. Alternatively, maximizing the window fixes it, ending the hang and causing the game to behave normally (and stopping the repeated logs), but still with significantly higher CPU usage than it exhibits when maximized with SDL classic (~40% vs ~7% on the title screen on my machine). Modifying the window size further seems to have unpredictable results, with the game alternately hanging and behaving normally. It hung while maximized during my experimentation, which suggests window size isn't the only factor.

Running the game with `SDL_VIDEODRIVER=wayland` had similar results, showing a black window and using 100% of one CPU while producing an identical log to stdout. However, the window had no decorations, and each time I maximized it with a keyboard shortcut, it immediately restored itself to its previous size and continued to hang. I had to `kill -9` it.

[...]

PR #129, opened 2021-08-14
Don't recreate the window on resize, even in windowed mode

sulix:
Resizable SDL 1.2 windows are really annoying to use when the window gets repeatedly recreated, as (at least on some window managers) this resets the resize operation.

This follows on from bug #127, as Dwarf Fortress hits this issue, being a resizable window which calls ``SDL_SetVideoMode()`` on every ``SDL_VIDEORESIZE`` event. With this change, resizing now seems to work identically to the SDL 1.2 version.

I'm not sure why we explicitly disabled this, though, so maybe there's a regression somewhere I don't know about. If so, maybe this is something we only do for ``SDL_RESIZABLE`` windows, or something?

### DIFF
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -4324,8 +4324,6 @@ SDL_SetVideoMode(int width, int height, int bpp, Uint32 flags12)
         EndVidModeCreate();  /* rebuild the window if moving to/from a GL context */
     } else if (VideoSurface12 && (VideoSurface12->surface20->format->format != appfmt)) {
         EndVidModeCreate();  /* rebuild the window if changing pixel format */
-    } else if (VideoSurface12 && (VideoSurface12->w != width || VideoSurface12->h != height) && ((flags12 & SDL12_FULLSCREEN) == 0)) {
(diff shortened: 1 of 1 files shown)


sezero:
@icculus ?

icculus:
I have to check the ramifications here, but the likely reason for this is that SDL 1.2 requires the app to call SDL_SetVideoMode when the window is resized, and calling SDL_SetVideoMode is expected to destroy the existing OpenGL context.

sulix:
In theory, I think we should still be recreating the OpenGL context (at least, for OpenGL windows) even with this change. The difference is that we're not recreating the whole window.

The next ``else if (VideoGLContext20)`` line should take care of that.

That being said, I haven't tested that path: I don't think I have any OpenGL-using SDL 1.2 apps with resizable windows (which handle VIDEORESIZE messages) lying around on this machine.

sezero:
What is the status of this patch?

icculus:
That patch is probably correct. Going to merge it and we'll see if any GL apps complain.

Issue #135, opened 2021-11-03
Non-US Keyboard layouts not supported


DanielGibson:
Just to make sure this is in a proper issue and not just in comments on a closed PR: SDL1.2 *did* support non-US layouts. Maybe not as well as SDL2 does, but contrary to the assumptions in #97 it did **not** just use scancodes.

(Pasting my comment from #97:)

For example, look at SDL1.2's `X11_TranslateKeycode()`: https://github.com/libsdl-org/SDL-1.2/blob/main/src/video/x11/SDL_x11events.c#L1155
It translates the X11 KeyCode (which is kinda like a SDL scancode) to a X11 KeySym (which is supposed to be the thing printed on the keyboard's keycaps, i.e. like SDL2 SDL_KeyCode, see https://tronche.com/gui/x/xlib/input/keyboard-encoding.html) and then translates that to a `SDLKey` which is used in `SDL_keysym::sym` (see https://github.com/libsdl-org/SDL-1.2/blob/main/src/video/x11/SDL_x11events.c#L606 ).

So `SDLKey` indeed is based on the name of (printed on) the key and not the scancode.

So https://github.com/libsdl-org/sdl12-compat/blob/36a09e450a007c86ed0f91812fbbba0d783be120/src/SDL12_compat.c#L2564 should be turned to `#if 1` and *maybe* some additional work could be done to support more (non-ASCII) keys; I'm not sure how exactly those (e.g. German Umlaut keys) were handled in SDL1.2.

I hope I'll get around to look into this a bit, but I don't know when so I can't promise anything right now.

sulix:
So, I did test this a bit with different keyboard layouts, and the conclusion I came to was that, while SDL 1.2 does support keyboard layouts, it's very often broken (it didn't work at all for me if there are multiple configured layouts), and more applications broke with it than without it. So I've sort-of been treating SDL 1.2 as not supporting keyboard layouts (except for the "unicode" value), and the X11 backend's support for them is more of a bug.

Similarly, my attempt at testing this on windows also had SDL 1.2 not take keyboard layout into account. Indeed, there appears to be some explicit code to force the US layout:
https://github.com/libsdl-org/SDL-1.2/blob/main/src/video/windib/SDL_dibevents.c#L547

[...]

PR #136, opened 2021-11-05
Add SDL12COMPAT_USE_KEYBOARD_LAYOUT hint

sulix:
tl;dr: This change makes the behaviour change in #97 optional, and configured at runtime with the new ``SDL12COMPAT_USE_KEYBOARD_LAYOUT`` environment variable. The default is the current behaviour of always using the scancodes/US layout, and the user's current layout can be used by setting said variable to '1'.

SDL 1.2 seems to only account for (non-en_US) keyboard layouts on some platforms and configurations. In particular, the KeySym may contain the translated key, or may contain a "scancode" which is layout-independent.

In particular, on X11 the configured layout is used, but only (it seems) if it is the only configured layout. (On systems where multiple layouts are configured, the US layout is used regardless of the configured one.) Similarly, the windib backend deliberately forces the US layout in order to match the DirectX backend, which only provides scancodes.

Since this behaviour is very inconsistant between different configurations, this patch provides a way of configuring which is used at runtime:

If the ``SDL12COMPAT_USE_KEYBOARD_LAYOUT`` environment variable is set to a nonzero value, sdl12compat will translate keyboard input according to the current layout. This matches SDL 1.2 on X11 with only one configured
layout. Otherwise (the default), sdl12compat will use SDL2 scancodes for keyboard input, essentially forcing all input to the US layout. This matches SDL 1.2 on Windows, and on X11 with multiple configured layouts.

Regardless of the value of this environment variable, the 'unicode' value will be affected by keyboard layout.

For discussion and more details see:
- [sdl12compat issue #135: Non-US Keyboard layouts not supported](https://github.com/libsdl-org/sdl12-compat/issues/135)
- [sdl12compat PR #97: Use Scancodes instead of Keycodes](https://github.com/libsdl-org/sdl12-compat/pull/97)
- [SDL 1.2 issue #95: windib returns locale dependent SDLK_ codes](https://github.com/libsdl-org/SDL-1.2/issues/95)

### DIFF
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -923,6 +923,7 @@ static char *WindowTitle = NULL;
 static char *WindowIconTitle = NULL;
 static SDL_Surface *VideoIcon20 = NULL;
 static int EnabledUnicode = 0;
+static SDL_bool TranslateKeyboardLayout = SDL_FALSE;
 static int VideoDisplayIndex = 0;
 static SDL_bool SupportSysWM = SDL_FALSE;
 static SDL_bool CDRomInit = SDL_FALSE;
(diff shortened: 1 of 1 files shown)


src/SDL12_compat.c
```diff
@@ -3108,7 +3112,11 @@ EventFilter20to12(void *data, SDL_Event *event20)
                 return 1;  /* ignore 2.0-style key repeat events */
             }
 
-            PendingKeydownEvent.key.keysym.sym = KeysymFromSDL2(event20);
+            if (TranslateKeyboardLayout) {
+                PendingKeydownEvent.key.keysym.sym = Keysym20to12(event20->key.keysym.sym);
+            } else {
+		        PendingKeydownEvent.key.keysym.sym = Scancode20toKeysym12(event20->key.keysym.scancode);
```
slouken, line 3118: Can you fix the whitespace here? Otherwise it looks good!

PR #137, opened 2021-11-07
Make SDL12COMPAT_USE_KEYBOARD_LAYOUT's default platform-specific.

sulix:
It looks like (at least out of the top three of Windows, Mac, and Linux), only Windows forces the use of the US layout. Mac (at least from a cursory reading of the code), translates according to the keyboard layout, as does X11 (albeit with some bugs).

In this case, set the default to disable layout translation only on Win32, and have it enabled by default on other platforms.

This does break DOSBox, but only when "usescancodes = false" (or potentially if keyboard layouts changed some of the non-printable character keys which would break out scancode fallback). Since this is not the default, hopefully people won't encounter it…?

### DIFF
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -923,7 +923,17 @@ static char *WindowTitle = NULL;
 static char *WindowIconTitle = NULL;
 static SDL_Surface *VideoIcon20 = NULL;
 static int EnabledUnicode = 0;
+/* Windows SDL1.2 never uses translated keyboard layouts for compatibility with
+DirectInput, which didn't support them. Other platforms (MacOS, Linux) seem to,
+but with varying levels of bugginess. So default to Translated Layouts on
+all non-Windows platforms (but have an option to change this, as many apps are
+buggy with non-US layouts, or provide their own keyboard layout translation,
+such as DOSBox. */
+#if !defined(_WIN32)
+static SDL_bool TranslateKeyboardLayout = SDL_TRUE;
+#else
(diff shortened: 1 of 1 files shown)


sezero:
> This does break DOSBox, but only when "usescancodes = false" (or potentially if keyboard layouts changed some of the non-printable character keys which would break out scancode fallback).

Does real SDL1.2 not fail with "usescancodes = false" or the other cases you describe?

sulix:
> > This does break DOSBox, but only when "usescancodes = false" (or potentially if keyboard layouts changed some of the non-printable character keys which would break out scancode fallback).
> 
> Does real SDL1.2 not fail with "usescancodes = false" or the other cases you describe?

For the "usescancodes = false" case, I suspect it will, though given real SDL 1.2 doesn't translate keyboard layouts on my setup — due, it seems, to having multiple configured layouts — I can't reproduce it here.

Potentially for some other cases, no. On Linux, the SDL 1.2 scancodes are different depending on whether or not evdev is used. DOSBox tries to detect this, and so handles scancodes differently. In order to avoid having to have a bunch of weird evdev detection code in sdl12compat, it currently returns a scancode of 0 for any key which is different between evdev/non-evdev cases, which causes DOSBox to fall back to the "usescancodes = false" path, which (as mentioned) could be broken. Real SDL 1.2 won't fallback here.

The only keys which should trigger that are the arrow keys and Insert/Home/PgUp/etc block — none of the keyboard layouts I tried changed those, so wouldn't be affected. So I think this is a pretty rare issue.

sezero:
Then let's merge this and handle any possible fallout later.

PR #139, opened 2021-11-07
Improve non-US layouts

DanielGibson:
* Fix scancode for they key to the right ofleft shift on ISO layouts
* Support SDLK_WORLD_* keys


DanielGibson:
Would be nice if someone with a Mac and an ISO keyboard layout (the kind of keyboards that have a short left shift and an additional key between that one and SDL_SCANCODE_Z) could verify if the scancodes used for that key are sane.
In fact I'd like to know:
* The SDL2 scancodes 
* The SDL1.2 scancodes from a "real" SDL1.2
* The SDL1.2 scancodes with SDL12-compat and this patch
**on Macs with ISO layout**, generated for:
* That key (the one right of Left Shift)
* the key that should be SDL_SCANCODE_GRAVE, which is the one between Esc, Tab and 1.

PR #140, opened 2021-11-07
checkkeys test: Improve printing of Unicode-Chars

DanielGibson:
.. at least on platforms that use UTF-8, like modern Linux (Windows should work as well as before).

Also unified the output to print sth like "'Ä' (0x00C4)", i.e. both the actual char and the unicode constant, on all platforms.

Before these changes, on Linux it would print garbage or nothing for the unicode character, if it wasn't from the ASCII range


sulix:
Thanks — I can confirm this works much better (at least on Linux) than the previous version.

I didn't test the windows codepath: it is possibly still a little bit fragile (I suspect it wouldn't handle > 16bit characters, or non-unicode systems with non-Latin-1 codepages), but it's not a regression, at least, and those cases should be pretty rare.

Issue #143, opened 2021-11-21
FIXME cleanup


icculus:
> There are a bunch of FIXME()s that it'd be nice to get to: even if we don't get to fix them, it'd be nice to have some of them only trigger when they're more likely to indicate an actual problem with the running application. Now that FIXMEs are disabled by default, it's not as confusing as it could be, though.

_Originally posted by @sulix in https://github.com/libsdl-org/sdl12-compat/issues/53#issuecomment-972897416_

icculus:
Putting this in the 1.2.50 milestone, but once we've made a pass over all the FIXMEs (many of them extremely aggressive wishlist stuff that may _never_ get changed), we should close this bug, even if most of the FIXMEs remain.

sezero:
PR #145 is in. Anything left?

sezero:
And https://github.com/libsdl-org/sdl12-compat/commit/4601def7935f8800b0b32750db84aa66823579b8 added two new FIXMEs. Heh he :)

sezero:
There is also an `/* !!! IMPLEMENT_ME X11_KeyToUnicode ? */` note at the top...

icculus:
IIRC, there was a Loki game (I forget which), which uses X11_KeyToUnicode, which was an internal function in the X11 backend. At the time, it was non-trivial to mark non-static symbols in shared libraries as private, so Loki used one of them they shouldn't have.

icculus:
...I want to say it was Heavy Gear II, but I could be wrong.

icculus:
Actually, it appears to be the Torque Engine having this issue in more modern times (although perhaps they inherited this from Tribes 2?)...this means Space Pirates and Zombies, on Steam will need this symbol exported, if nothing else.

We should open a bug for this.

icculus:
> FIXME("there is never a parachute in SDL2, should we catch segfaults ourselves?"); — Not in v1, I think.

So the primary reason for the parachute was to reset the video mode if an app crashes, so you didn't end up with a 320x200 desktop or whatever on X11...but this is not an issue with FULLSCREEN_DESKTOP, so the only case where we could hit this now is if the user has forced off OpenGL Scaling in a GL app.

(Actually, this might have been an XVidMode issue we don't get with XRandR...?)

icculus:
> FIXME("Respect 1.2 environment variables for defining format here.");
>
>  * This comes from SDL_OpenAudio()
[...]

PR #145, opened 2021-11-22
Bunch of FIXME-related fixes

sulix:
[I'm happy to split one or all of these out into separate PRs if some are more controversial than others.]

This is a bunch of changes related to various FIXMEs in the sdl12-compat code, as noted in #143. This change includes:

- ``glRenderbufferStorate()`` is a part of core GL 3, so get rid of the fixme reminding us to check.
- ``SDL_VIDEO_FULLSCREEN_DISPLAY`` was only partly supported, complete it.
- Make the timing of presentations in UpdateRects and when handling YUV overlays refresh-rate dependent.
- Add a very basic ``SDL_EnableKeyRepeat()`` implementation which ignores the delay/interval and just uses SDL2's repeat field.

(Largely in order from least to most controversial.) The only real issues I can imagine are:
- The ``GetEnvironmentWindowPosition()`` function now sets x and y unconditionally, so intialising them to ``SDL_WINDOWPOS_UNDEFINED`` in ``SDL_SetVideoMode()`` is now useless.
- ``GetDesiredMillisecondsPerFrame()`` could be more accurate in windowed-mode if the window is dragged between different screens.
- ``GetDesiredMillisecondsPerFrame()`` has (obviously) millisecond resolution, so will not be able to accurately represent, e.g, 60Hz displays. (But it's better than the hardcoded 15).
- ``SDL_EnableKeyRepeat()`` doesn't support the ``delay``/``interval`` parameters properly. Even when key repeats are disabled, we still get KEYDOWN events via the SDL2 TEXTINPUT events (though this existed prior to this change).

### DIFF
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -923,6 +923,7 @@ static char *WindowTitle = NULL;
 static char *WindowIconTitle = NULL;
 static SDL_Surface *VideoIcon20 = NULL;
 static int EnabledUnicode = 0;
+static SDL_bool EnabledKeyRepeat = SDL_TRUE;
 /* Windows SDL1.2 never uses translated keyboard layouts for compatibility with
 DirectInput, which didn't support them. Other platforms (MacOS, Linux) seem to,
 but with varying levels of bugginess. So default to Translated Layouts on
@@ -3785,7 +3786,7 @@ EventFilter20to12(void *data, SDL_Event *event20)
 
         case SDL_KEYDOWN:
             FlushPendingKeydownEvent(0);
-            if (event20->key.repeat) {
+            if (event20->key.repeat && !EnabledKeyRepeat) {
                 return 1;  /* ignore 2.0-style key repeat events */
             }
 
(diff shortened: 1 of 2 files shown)


src/SDL12_compat.c
```diff
@@ -5581,6 +5581,24 @@ UpdateRect12to20(SDL12_Surface *surface12, const SDL12_Rect *rect12, SDL_Rect *r
     }
 }
 
+/* For manual throttling of screen updates. */
+static int
+GetDesiredMillisecondsPerFrame()
+{
+    SDL_DisplayMode mode;
+    if (VideoSurface12->flags & SDL12_FULLSCREEN) {
+       if (!SDL20_GetWindowDisplayMode(VideoWindow20, &mode)) {
```
slouken, line 5591: For SDL functions that return 0 or an error, you should check == 0, not !func.

src/SDL12_compat.c
```diff
@@ -5581,6 +5581,24 @@ UpdateRect12to20(SDL12_Surface *surface12, const SDL12_Rect *rect12, SDL_Rect *r
...
+           return 1000 / mode.refresh_rate;
```
slouken, line 5592: Need to check mode.refresh_rate is not 0 here like you do below.

src/SDL12_compat.c
sulix: Done.

src/SDL12_compat.c
sulix: Whoops! Fixed.

Issue #168, opened 2022-03-14
Awesomenauts (Steam, free-to-play): black window, no titlebar or borders


smcv:
* Debian 10 host system, AMD GPU, Mesa
* run Steam with a modified Steam Runtime that contains sdl12-compat 1.2.52 instead of "real" SDL 1.2
* launch https://store.steampowered.com/app/204300/Awesomenauts__the_2D_moba/

Expected result: normal gameplay

Actual result: a black window with no titlebar or borders (but the game is running and sounds still play)

slouken:
Ryan, can you take a look at this?

icculus:
Checking now, but I assume it's going to be a framebuffer object issue, and disabling OpenGL scaling will fix it. One moment.

smcv:
I should have mentioned that I'm testing under GNOME in X11 mode, in case it matters.

Launchign with `SDL12COMPAT_OPENGL_SCALING=0 %command%` changes the result to: instead of a black window, I get a window whose contents are a snapshot of what was underneath the window at the time it was launched (in my case, that's the Steam UI).

icculus:
So the titlebar/border issue is that Awesomenauts is calling SDL_SetVideoMode(width of desktop, height of desktop, 32, SDL_NOFRAME)...what we might call "borderless fullscreen" (or whatever) in some games.

Should we detect this and create the SDL2 window with SDL_WINDOW_FULLSCREEN_DESKTOP...? It seems like an obvious thing to do that likely has some non-obvious corner cases.

Still checking into the black screen issue.

icculus:
Adding that (and a switch to turn it off, if it makes something misbehave, like we do for GL scaling).

smcv:
I'm using vanilla GNOME on this particular test system (no dock, just a vaguely Mac-like top bar), and games that go into "borderless fullscreen" are usually genuinely full-screen here (they end up "in front of" the top bar). My screen is 1920x1080.

I had Awesomenauts set to 1024x768 borderless fullscreen (sorry, I don't remember whether that was the default or something I did deliberately). With sdl12-compat, the result was a 1024x768 borderless window. With "real" SDL 1.2, the result was fullscreen, but the game's graphics were only drawn into the top left 1024x768 region, with the rest of the screen black.

[...]

PR #171, opened 2022-03-17
Implement an application-compatibility "quirks" system (and hacky workaround for Awesomenauts)

sulix:
This follows the discussion in #168 and introduces a "quirks" table which allows application-specific workarounds to be implemented.

Basically, it works by adding an ``SDL12Compat_GetHint()`` function to replace the various ``getenv`` calls around the place, and makes that check an array of (executable name, hint name, hint value) tuples.

The list of quirks is platform-specific, and keyed on the executable filename (hence there being duplicates for different architectures).

I've also added a workaround for the Awesomenauts issue in bug #168, though it's a bit hacky.

I've also not added, e.g., Dynamite Jack, whose executable name is — I think — just "main". Things like driconf and vkd3d-proton's system support fancier ways of selecting these (with wildcards, or executable hashes, etc), but this should do for now.

There's a bunch of optimisation options still left on the table (places where we could cache things, or enforce the table being in alphabetical order to break out sooner, etc), and a bit of other ugliness, but it should work well enough for now. Please add entries as you find them!

Thanks to @sezero for the OS/2 implementation of ``GetExeName``. (I hope it works. :-))

### DIFF
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -968,6 +968,7 @@ static GLuint OpenGLLogicalScalingMultisampleColor = 0;
 static GLuint OpenGLLogicalScalingMultisampleDepth = 0;
 static GLuint OpenGLCurrentReadFBO = 0;
 static GLuint OpenGLCurrentDrawFBO = 0;
+static SDL_bool ForceGLSwapBufferContext = SDL_FALSE;
 
 /* !!! FIXME: need a mutex for the event queue. */
 #define SDL12_MAXEVENTS 128
@@ -1138,6 +1139,7 @@ static QuirkEntryType quirks[] = {
     /* Awesomenauts uses Cg, and does weird things with the GL context. */
     {"Awesomenauts.bin.x86", "SDL12COMPAT_OPENGL_SCALING", "0"},
     {"Awesomenauts.bin.x86", "SDL_VIDEODRIVER", "x11"},
+    {"Awesomenauts.bin.x86", "SDL12COMPAT_FORCE_GL_SWAPBUFFER_CONTEXT", "1"},
 
     /* Braid uses Cg, which uses glXGetProcAddress(). */
     {"braid", "SDL_VIDEODRIVER", "x11"},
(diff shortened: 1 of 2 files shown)


src/SDL12_compat.c
```diff
@@ -4958,13 +5080,12 @@ SDL_SetVideoMode(int width, int height, int bpp, Uint32 flags12)
            that binds our logical FBO instead of the window framebuffer, we can probably
            work with these apps, too. That's easy from SDL_GL_GetProcAddress, but we
            maybe need to export the symbol from here too, for those that link against
-           OpenGL directly. UT2004 is known to use FBOs with SDL 1.2, and I assume
+           OpenGL directly. UT2004 is known to use FBOs with SDL 1., NULL2, and I assume
```
sezero, line 5083: ???

src/SDL12_compat.c
```diff
@@ -6596,9 +6711,14 @@ DECLSPEC void SDLCALL
 SDL_GL_SwapBuffers(void)
 {
     if (VideoWindow20) {
+        /* Some applications, e.g. Awesomenauts, play with glXMakeCurrent() behind out backs and break SwapBuffers() */
+        if (SDL12Compat_GetHintBoolean("SDL12COMPAT_FORCE_GL_SWAPBUFFER_CONTEXT", SDL_FALSE)) {
```
sezero, line 6715: How cheap will this be? Can't cache in some state or flag?

sezero:
> Thanks to @sezero for the OS/2 implementation of `GetExeName`. (I hope it works. :-))

It should :) (It's the same code used in SDL_GetBasePath() in current SDL git.)

sezero:
What to do with other unices, e.g. freebsd?

What to do with macOS?

src/SDL12_compat.c
sulix: Whoops: I must have accidentally pasted that. Fixed.

src/SDL12_compat.c
sulix: It wasn't expensive enough that I noticed on my system, but I've cached it nevertheless, as it definitely could be more expensive on others.

sulix:
> What to do with other unices, e.g. freebsd?
> 
> What to do with macOS?

I haven't looked into to other unices or macOS, largely because I don't know them well enough (or have them set up to test) to be sure I'd got them right.

Do we know of anything for any other platforms which would need a quirk? I imagine the BSDs could be hit by the various glX issues, but most things would either be open-source or run under the Linux compatibility layer anyway.

For macOS, is there anything that needs a quirk now?

icculus:
(I'm not aware of any macOS quirks atm, but we can deal with that in main when they pop up.)

PR #174, opened 2022-03-20
Resolve relative paths when initialising CD audio

sulix:
Some games will ``chdir()`` in between initialising SDL and trying to play CD audio, which will break if the path passed to ``SDL12COMPAT_FAKE_CDROM_PATH`` is relative. Instead, resolve the relative path to an absolute path in ``InitializeCDSubsystem()``. It's still going to be a pain for the user if the game calls ``chdir()`` before even SDL is initialised, but it's better than nothing.

Implemented for (and tested on) Win32 and unix. Anything with ``realpath()`` should work.

One game which triggers this is the old Loki port of Civilization: Call to Power (which is technically SDL 1.1, admittedly).

### DIFF
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -7455,7 +7455,17 @@ InitializeCDSubsystem(void)
 
     cdpath = SDL12Compat_GetHint("SDL12COMPAT_FAKE_CDROM_PATH");
     if (cdpath) {
+        char abspath[SDL12_MAXPATH];
+        /* Some games chdir() after initialising, making relative paths break. */
+#ifdef WIN32
+        /* There is a _fullpath() function, but it needs libc, which we don't use on Win32. */
+        GetFullPathNameA(cdpath, SDL12_MAXPATH, abspath, NULL);
+        CDRomPath = SDL_strdup(abspath);
+#elif defined(__unix__)
+        CDRomPath = SDL_strdup(realpath(cdpath, abspath));
+#else
         CDRomPath = SDL_strdup(cdpath);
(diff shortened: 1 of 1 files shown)


src/SDL12_compat.c
```diff
@@ -7455,7 +7455,17 @@ InitializeCDSubsystem(void)
 
     cdpath = SDL12Compat_GetHint("SDL12COMPAT_FAKE_CDROM_PATH");
     if (cdpath) {
+        char abspath[SDL12_MAXPATH];
+        /* Some games chdir() after initialising, making relative paths break. */
+#ifdef WIN32
+        /* There is a _fullpath() function, but it needs libc, which we don't use on Win32. */
+        GetFullPathNameA(cdpath, SDL12_MAXPATH, abspath, NULL);
+        CDRomPath = SDL_strdup(abspath);
+#elif defined(__unix__)
```
sezero, line 7464: ` || defined(__APPLE__)`

sezero:
Do we really want this, though? Wouldn't it be easier to document that an abs path is recommended?

icculus:
I agree with @sezero, let's not do this.

sulix:
I guess what most people actually want is to have this set to SDL_GetBasePath(), or something relative to that.

But it all could be done in a wrapper script which resolves everything down to an absolute path if necessary, I suppose…

sezero:
Pushed https://github.com/libsdl-org/sdl12-compat/commit/85d1f7ab0e22cdd526eb2ebf6021f247d9f55a3e to doco

PR #179, opened 2022-04-01
BSD support: Don't assume presence of libdl on Unix

alarixnia:
Most non-Linux OSes have dlopen() in libc, and CMake provides
a convenient variable for making use of libdl conditional.


sezero:
Thanks

Issue #183, opened 2022-05-01
schismtracker crashes when resizing window


ghost:
terminal output:
```
sdl_getwminfo?
double free or corruption (!prev)
Aborted (core dumped)
```
sometimes i also get this:
`malloc(): invalid next size (unsorted)`

bt:
```
#0  0x00007ffff7bc334c in __pthread_kill_implementation () from /usr/lib/libc.so.6
#1  0x00007ffff7b764b8 in raise () from /usr/lib/libc.so.6
#2  0x00007ffff7b60534 in abort () from /usr/lib/libc.so.6
#3  0x00007ffff7bb7397 in __libc_message () from /usr/lib/libc.so.6
#4  0x00007ffff7bcd33c in malloc_printerr () from /usr/lib/libc.so.6
#5  0x00007ffff7bd05fc in _int_malloc () from /usr/lib/libc.so.6
#6  0x00007ffff7bd2489 in calloc () from /usr/lib/libc.so.6
#7  0x00007ffff5ad2803 in ?? () from /usr/lib/dri/crocus_dri.so
#8  0x00007ffff5ad8d00 in ?? () from /usr/lib/dri/crocus_dri.so
#9  0x00007ffff4c8ce6e in ?? () from /usr/lib/dri/crocus_dri.so
#10 0x00007ffff4c406cf in ?? () from /usr/lib/dri/crocus_dri.so
#11 0x00007ffff4c4ba89 in ?? () from /usr/lib/dri/crocus_dri.so
#12 0x00007ffff4c4bb79 in ?? () from /usr/lib/dri/crocus_dri.so
#13 0x00007ffff7998d39 in ?? () from /usr/lib/libSDL2-2.0.so.0
#14 0x00007ffff798f940 in ?? () from /usr/lib/libSDL2-2.0.so.0
#15 0x00007ffff7f84db2 in SDL_UpdateRects () from /usr/lib/libSDL-1.2.so.0
#16 0x00007ffff7f84fa8 in SDL_UpdateRect () from /usr/lib/libSDL-1.2.so.0
#17 0x00007ffff7f85033 in SDL_Flip () from /usr/lib/libSDL-1.2.so.0
#18 0x0000555555560797 in ?? ()
#19 0xffffffff00000000 in ?? ()
#20 0x0000000000000000 in ?? ()


```

sulix:
I can reproduce this here.

[...]

PR #184, opened 2022-05-02
Use SDL_PIXELFORMAT_RGB888 for 24-bit surfaces

sulix:
We're currently using SDL_PIXELFORMAT_RGB24, which results in
schismtracker being blue when resized (#183).

While it's possible that RGB888 is wrong on big-endian systems, it at
least matches what we do for 16-bit (RGB565) and 32-bit (XRGB8888)
formats, so if we're wrong, we'll at least be consistently wrong.

### DIFF
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -5231,7 +5231,7 @@ SDL_SetVideoMode(int width, int height, int bpp, Uint32 flags12)
     switch (bpp) {
         case  8: appfmt = SDL_PIXELFORMAT_INDEX8; break;
         case 16: appfmt = SDL_PIXELFORMAT_RGB565; FIXME("bgr instead of rgb?"); break;
-        case 24: appfmt = SDL_PIXELFORMAT_RGB24; FIXME("bgr instead of rgb?"); break;
+        case 24: appfmt = SDL_PIXELFORMAT_RGB888; FIXME("bgr instead of rgb?"); break;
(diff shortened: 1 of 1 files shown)


src/SDL12_compat.c
```diff
@@ -5231,7 +5231,7 @@ SDL_SetVideoMode(int width, int height, int bpp, Uint32 flags12)
     switch (bpp) {
         case  8: appfmt = SDL_PIXELFORMAT_INDEX8; break;
         case 16: appfmt = SDL_PIXELFORMAT_RGB565; FIXME("bgr instead of rgb?"); break;
-        case 24: appfmt = SDL_PIXELFORMAT_RGB24; FIXME("bgr instead of rgb?"); break;
+        case 24: appfmt = SDL_PIXELFORMAT_RGB888; FIXME("bgr instead of rgb?"); break;
```
slouken, line 5234: I don't think this is right. The app is asking for a 24-bit surface, and you're returning a 32-bit surface. If it ends up touching the pixels, then it will be very wrong if it's blindly assuming that SDL hasn't changed the format.

@icculus, am I missing something here?

icculus: I'd have to check, but maybe if asked for 24 bit color you got XRGB back from the X server in SDL 1.2...? It's certainly possible it worked out like that.

slouken: Yep, that's likely.

Issue #187, opened 2022-05-30
Reading key presses not working when CTRL key is held down


soruk42:
My use case is the Matrix Brandy BASIC interpreter, which uses SDL 1.2 (I have a plan to port it, though not sure yet how to deal with the way it writes pixels directly to the framebuffer...). So far most is working, but keypresses involving the CTRL key are not picked up.  This is easily demonstrated within Matrix Brandy, by issuing
`PRINT GET`
and pressing CTRL and any key, you should get that sequence ASCII code back, e.g. CTRL-Q should give 17. When run against the real SDL-1.2.15 this works perfectly, but against sdl12-compat no keypress is seen.

soruk42:
I've run some comparisons between SDL-1.2.15 and the latest asl12-compat from github, and it appears when the CTRL key is pressed they keysym.unicode struct member isn't completed.
SDL-1.2.15
'd'
keysym.unicode=100 keysym.sym=100 keysym.mod=0x0
SHIFT+'d'
keysym.unicode=68 keysym.sym=100 keysym.mod=0x1
CTRL+'d'
**keysym.unicode=4** keysym.sym=100 keysym.mod=0x40
ALT+'d'
keysym.unicode=100 keysym.sym=100 keysym.mod=0x100

sdl12-compat:
'd'
keysym.unicode=100 keysym.sym=100 keysym.mod=0x0
SHIFT+'d'
keysym.unicode=68 keysym.sym=100 keysym.mod=0x1
CTRL+'d'
**keysym.unicode=0** keysym.sym=100 keysym.mod=0x40
ALT+'d'
keysym.unicode=100 keysym.sym=100 keysym.mod=0x100

PR #189, opened 2022-07-28
Avoid crash if NULL is passed to SDL_RemoveTimer and add fallback for unicode when CTRL is pressed.

akb825:


icculus:
The patch is fine, but the FIXME comment needs to stay; the real SDL 1.2 won't crash with an invalid, non-NULL pointer here, but we will.

akb825:
I also encountered the issue noted in #187, which I included a fix for with this PR.

akb825:
> The patch is fine, but the FIXME comment needs to stay; the real SDL 1.2 won't crash with an invalid, non-NULL pointer here, but we will.

I see, I was only thinking of NULL when I removed the comment.

src/SDL12_compat.c
```diff
@@ -4056,7 +4056,16 @@ EventFilter20to12(void *data, SDL_Event *event20)
...
                     break;
                 default:
-                    /* not a supported control character */
+                    /* not a supported control character
+                       when CTRL is pressed, text events aren't sent so use fallback for unicode */
+                    if (PendingKeydownEvent.key.keysym.mod & KMOD_CTRL) {
+                        const char *keyName = SDL_GetKeyName(PendingKeydownEvent.key.keysym.sym);
+                        /* use key name as unicode if it's a single character */
+                        if (keyName[0] && !keyName[1])
+                            FlushPendingKeydownEvent(keyName[0]);
```
akb825, line 4065: I'm not familiar with the SDL 1.2 API, should this be adjusted if shift is pressed? Letters should be easy enough with toupper(), but not sure about symbols since they could be different based on the keyboard layout...

PR #195, opened 2022-08-28
Check that HAVE_* macros are true, not just defined

jmroot:
Other build systems can define these but to 0 when the feature is
unavailable, and include this header.

### DIFF
--- a/include/SDL/SDL_stdinc.h
+++ b/include/SDL/SDL_stdinc.h
@@ -77,43 +77,43 @@ typedef enum SDL_DUMMY_ENUM
 
 SDL_COMPILE_TIME_ASSERT(enum, sizeof(SDL_DUMMY_ENUM) == sizeof(int));
 
-#ifdef HAVE_SYS_TYPES_H
+#if HAVE_SYS_TYPES_H
 #include <sys/types.h>
 #endif
-#ifdef HAVE_STDIO_H
+#if HAVE_STDIO_H
 #include <stdio.h>
 #endif
-#ifdef HAVE_STDLIB_H
+#if HAVE_STDLIB_H
 #include <stdlib.h>
 #endif
-#ifdef HAVE_MALLOC_H
+#if HAVE_MALLOC_H
 #include <malloc.h>
 #endif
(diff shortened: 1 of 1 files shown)


jmroot:
If we really wanted to avoid collisions, we would use a project-specific prefix on the macro names, but I don't know if that would affect backward compatibility.

sezero:
> Other build systems

Which ones?

>  can define these but to 0 when the feature is unavailable

Wouldn't that be a buggy behavior???

jmroot:
MPlayer was subject of the report that made me aware of this. I don't think it's necessarily buggy to use the convention of `#define HAVE_FOO 0` meaning you don't have feature "foo" and `#define HAVE_FOO 1` meaning you do. In any case, this is fixing a break in backward compatibility, because defining `HAVE_MALLOC_H` to 0 causes the build to fail with sdl12-compat when it succeeds with the real SDL 1.2.

sezero:
Real SDL-1.2 also checks malloc.h like `# elif defined(HAVE_MALLOC_H)`:
https://github.com/libsdl-org/SDL-1.2/blob/main/include/SDL_stdinc.h#L46

I don't think that we want this patch. @icculus?

jmroot:
Switching to doing it that way would also work I guess; the -compat header currently always checks `HAVE_MALLOC_H`, not just if `HAVE_STDLIB_H` is undefined.

slouken:
I'm closing this for now. We can revisit #if vs #ifdef in SDL3

sezero:
> We can revisit #if vs #ifdef in SDL3

The `#if HAVE_xxx` version will result in `-Wundef` warnings

DanielGibson:
> If we really wanted to avoid collisions, we would use a project-specific prefix on the macro names

This is something that should definitely be done at least in SDL3, maybe even SDL2 if it doesn't break anyones code.
Because I have seen clashes (or more specifically warnings from the compiler about redefinitions) with the headers from other libs (libjpeg maybe?) that also `#define HAVE_STDLIB_H` or similar.
`SDL_HAVE_STDLIB_H` and friends should prevent that kind of conflict

PR #196, opened 2022-08-29
Make include logic closer to real SDL 1.2

jmroot:
This should offer better compatibility in some cases.

This PR is an alternative to #195. I think that one is probably better practice in general, but this one is taking more of a bug-for-bug compatibility approach, as may better fit the goals of sdl12-compat.

### DIFF
--- a/include/SDL/SDL_stdinc.h
+++ b/include/SDL/SDL_stdinc.h
@@ -83,10 +83,9 @@ SDL_COMPILE_TIME_ASSERT(enum, sizeof(SDL_DUMMY_ENUM) == sizeof(int));
 #ifdef HAVE_STDIO_H
 #include <stdio.h>
 #endif
-#ifdef HAVE_STDLIB_H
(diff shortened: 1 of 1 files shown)


include/SDL/SDL_stdinc.h
```diff
@@ -83,10 +83,9 @@ SDL_COMPILE_TIME_ASSERT(enum, sizeof(SDL_DUMMY_ENUM) == sizeof(int));
 #ifdef HAVE_STDIO_H
 #include <stdio.h>
 #endif
-#ifdef HAVE_STDLIB_H
+#if defined(HAVE_STDLIB_H)
```
slouken, line 86: Aren't these two equivalent?

jmroot: Yes. The important part is the use of an `#elif` instead of a second, separate `#if`.

slouken:
Okay, yes, this matches the original SDL-1.2, so it looks good to me.

PR #221, opened 2022-09-25
Clamp mouse coordinates to the backbuffer size for non-OpenGL

sulix:
When we're logically scaling, and preserving the aspect ratio (so are letterboxing or pillarboxing), the scaled mouse coordinates may be outside the application window (backbuffer). These can even be negative, which get underflowed to very large positive values, as SDL 1.2 uses unsigned coordinates here.

We were already clamping these values for OpenGL logical scaling, which we do ourselves, but the event filter used with SDL_Renderer doesn't, so we'll manually clamp these all the time.

The issue could be reproduced in Civilization: Call To Power, by starting the game windowed 1024×768 (or another 4:3 resolution), maximising the window (which, if you have a widescreen display, will pillarbox), and starting a game. When pushing the mouse to the left edge to scroll, you have to very precisely stay within the window. If the mouse ends up in the pillarbox region, the game map will scroll the other way, as the negative mouse coordinate will underflow, and the game will treat that as scrolling to the right.

### DIFF
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -4512,6 +4512,9 @@ EventFilter20to12(void *data, SDL_Event *event20)
             event12.motion.which = (Uint8) event20->motion.which;
             event12.motion.state = event20->motion.state;
             AdjustOpenGLLogicalScalingPoint(&event20->motion.x, &event20->motion.y);
+            /* Clamp the absolute position to the window dimensions. */
+            event20->motion.x = SDL_max(SDL_min(event20->motion.x, VideoSurface12->w), 0);
+            event20->motion.y = SDL_max(SDL_min(event20->motion.y, VideoSurface12->h), 0);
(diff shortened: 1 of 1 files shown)


icculus:
@slouken this should be fixed in SDL2's scaling code, too, right?

sulix:
FWIW, I don't think this is as obviously a bug in SDL2, for two reasons:
1. SDL2 applications are opting-in to scaling, and if they use it, they should be expecting the existing behaviour of out-of-bounds coordinates.
2. SDL2 mouse coordinates are signed, so the whole integer-underflow issue, which is usually what actually causes issues, isn't a factor.

slouken:
We also have the behavior in SDL now where pressing the mouse button will capture the mouse, and if you move the mouse beyond the bounds of the window, you'll still get mouse events with coordinates outside the window.

PR #224, opened 2022-09-27
cmake: make sure all libraries in CMAKE_DL_LIBS are added with -l prefix to .pc file

### LINKED ISSUE #222: building with -DSTATICBUILD=true mis-generates Libs.private in pkgconfig
Using the 1.2.56 release or git main, this is what is generated in the pkgconfig file.  sdl-config is also mis-generated:

```# sdl12_compat pkg-config source file

prefix=/usr/local
exec_prefix=${prefix}
libdir=/usr/local/lib64
includedir=/usr/local/include

Name: sdl12_compat
Description: An SDL-1.2 compatibility layer that uses SDL 2.0 behind the scenes.
Version: 1.2.56
Provides: sdl = 1.2.56
Libs: -L${libdir}  -lSDL
Libs.private: -L${libdir} -lSDL dl    # <--- WTF?
Cflags: -I${includedir}/SDL -D_GNU_SOURCE=1 -D_REENTRANT
```
Needless to say, this causes problems with anything trying to statically link against sdl12-compat!

I am running Fedora 36, which currently uses cmake 3.22.2 and gcc 12.2.1.

madebr:
- find SDL2 through [`find_package`](https://cmake.org/cmake/help/latest/command/find_package.html)
- prepend all values in `CMAKE_DL_LIBS` with `-l`, for use in the `.pc` file.
  The CMake script unconditionally uses `dl`, even if `dlopen` is available in the c library.
  This is also what's documented in the [dlopen manual](https://man7.org/linux/man-pages/man3/dlopen.3.html).

Fixes #222

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,7 @@ option(SDL12DEVEL "Enable installing SDL-1.2 development headers" ON)
 option(STATICDEVEL "Enable installing static link library" OFF)
 
 if(STATICDEVEL AND NOT (CMAKE_SYSTEM_NAME MATCHES "Linux"))
-  MESSAGE(FATAL_ERROR "Static builds are only supported on Linux.")
+  message(FATAL_ERROR "Static builds are only supported on Linux.")
 endif()
 
 list(APPEND CMAKE_MODULE_PATH
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,7 +102,7 @@ elseif(WIN32)
     set_target_properties(SDL PROPERTIES
         VERSION "${PROJECT_VERSION}"
         SOVERSION "0"
-      OUTPUT_NAME "SDL")
+        OUTPUT_NAME "SDL")
 elseif(OS2)
     set_target_properties(SDL PROPERTIES COMPILE_DEFINITIONS "BUILD_SDL") # for DECLSPEC
     set_target_properties(SDL PROPERTIES
@@ -113,7 +113,7 @@ else()
     set_target_properties(SDL PROPERTIES
         VERSION "${PROJECT_VERSION}"
         SOVERSION "0"
-      OUTPUT_NAME "SDL")
+        OUTPUT_NAME "SDL")
 endif()
 
 if(MINGW)
@@ -262,7 +262,10 @@ if(SDL12DEVEL)
       set(SDL_CFLAGS "-D_GNU_SOURCE=1 -D_REENTRANT")
       set(SDL_RLD_FLAGS "")  # !!! FIXME: this forces rpath, which we might want?
       set(SDL_LIBS "-lSDL")
-      set(SDL_STATIC_LIBS ${CMAKE_DL_LIBS})
+      set(SDL_STATIC_LIBS "")
+      foreach(lib ${CMAKE_DL_LIBS})
+          set(SDL_STATIC_LIBS "-l${lib}")
+      endforeach()
       if(NOT STATICDEVEL)
         set(SDL_STATIC_LIBS "")
       endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,11 @@ if(STATICDEVEL AND NOT (CMAKE_SYSTEM_NAME MATCHES "Linux"))
   MESSAGE(FATAL_ERROR "Static builds are only supported on Linux.")
 endif()
 
+include(CheckCSourceCompiles)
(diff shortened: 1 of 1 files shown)


sezero:
Line 311 of original file: `target_link_libraries(SDL-static PRIVATE ${CMAKE_DL_LIBS})`
should be removed I guess? No sense in setting link libs for a static archive.

CMakeLists.txt
```diff
@@ -262,7 +262,10 @@ if(SDL12DEVEL)
       set(SDL_CFLAGS "-D_GNU_SOURCE=1 -D_REENTRANT")
       set(SDL_RLD_FLAGS "")  # !!! FIXME: this forces rpath, which we might want?
       set(SDL_LIBS "-lSDL")
-      set(SDL_STATIC_LIBS ${CMAKE_DL_LIBS})
+      set(SDL_STATIC_LIBS "")
+      foreach(lib ${CMAKE_DL_LIBS})
+          set(SDL_STATIC_LIBS "-l${lib}")
```
sezero, line 267: This should check whether `${CMAKE_DL_LIBS})` is an empty string
which cmake can very well do.

madebr:
> Line 311 of original file: `target_link_libraries(SDL-static PRIVATE ${CMAKE_DL_LIBS})` should be removed I guess? No sense in setting link libs for a static archive.

No, it must be kept. `dl` will be added to any library/executable linking to `SDL-static`.

CMakeLists.txt
madebr: It does not. `foreach` ignores empty values:
```cmake
cmake_minimum_required(VERSION 3.0)
project(a C)
set(A "1;;;2;;;;3")
foreach(a ${A})
  message("a ${a}")
endforeach()
```
prints
```
a 1
a 2
a 3
```

This is probably done because a lot of cmake code does: `list(SOME_VALUE APPEND ${SOME_OPTIONAL_LIBRARY})`.

sezero: Or, does `foreach` handles that by itself??

madebr: To be honest, it is [not documented](https://cmake.org/cmake/help/latest/command/foreach.html).
So i'm not 100% sure about whether this is documented behavior.

sezero:
> > Line 311 of original file: `target_link_libraries(SDL-static PRIVATE ${CMAKE_DL_LIBS})` should be removed I guess? No sense in setting link libs for a static archive.
> 
> No, it must be kept. `dl` will be added to any library/executable linking to `SDL-static`.

No one in sdl12-compat links to static sdl12-compat? Or, are we talking about different things?

madebr:
> > > Line 311 of original file: `target_link_libraries(SDL-static PRIVATE ${CMAKE_DL_LIBS})` should be removed I guess? No sense in setting link libs for a static archive.
> > 
> > 
> > No, it must be kept. `dl` will be added to any library/executable linking to `SDL-static`.
> 
> No one in sdl12-compat links to static sdl12-compat? Or, are we talking about different things?

These are there for people vendoring sdl12-compat and for exported targets.

(exported targets are the targets created in `lib/cmake/SDL2/...` when installing SDL2)

sezero:
> These are there for people vendoring sdl12-compat and for exported targets.
> 
> (exported targets are the targets created in `lib/cmake/SDL2/...` when installing SDL2)

OK, I give up :/

CMakeLists.txt
sezero: OK, will you keep it as is or you want to make changes before merging?

CMakeLists.txt
```diff
@@ -48,7 +52,10 @@ set(SDL12COMPAT_SRCS
 )
 add_library(SDL SHARED ${SDL12COMPAT_SRCS})
 
-include("cmake/modules/FindSDL2.cmake")
```
sezero, line 51: I guess we can remove the module file itself after this?

madebr: No, it's still used through `find_package(SDL2)`.
Using `find_package(SDL2)` allows using an installed `SDL2Config.cmake`.
We added some level of indirection here.

CMakeLists.txt
sezero: > OK, will you keep it as is or you want to make changes before merging?

No changes needed? Shall I merge?

madebr: Okay by me!

sezero: OK then.

PR #227, opened 2022-10-03
README: Mention the ALLOW_THREADED_ quirks

smcv:


icculus:
Whoops, I forgot to do this, so thanks for writing that up!

Issue #217, opened 2022-09-23
macOS dylib doesn't explicitly load SDL 2.0 dylib, causing dylibbundler failures


Kreeblah:
I noticed when building [a project](https://github.com/Kreeblah/MazeOfGalious) against sdl12-compat instead of SDL 1.2 that it failed to run, stating that it couldn't find SDL 2.0.

After some investigation, I found that the step that causes it to fail to load is when [dylibbundler](https://github.com/auriamg/macdylibbundler) is run against it, which packs dependent shared libraries into a given app bundle, and rewrites their paths to create a self-contained bundle for distribution.  The reason it's failing is because it's using `otool -l` to look through each library to determine where there are `LC_LOAD_DYLIB` calls, and includes those, and the build process here doesn't link SDL 2.0 that way.

I thought this would be as easy as adding a `target_link_libraries(SDL PRIVATE ${SDL2_LIBRARIES})` to `CMakeFiles.txt` around line 83, but that seems to add `-framework -lCocoa` as well (which obviously doesn't exist as a framework).  I'm not sure where that `-l` prefix is coming from on the Cocoa framework name.

Hard-coding my SDL 2.0 library paths and the literal string `-framework Cocoa` using `target_link_libraries` statements did let it build, and I was able to verify that `dylibbundler` works with the resulting dylib, so linking the library seems to be a solution.  I'm just not quite sure where CMake is getting that `-l` prefix from on the Cocoa framework.

Is there any chance this could be looked at?

sezero:
You really shouldn't link to SDL2: sdl12-compat dlopen's it at runtime
and scans its presence at multiple locations:
https://github.com/libsdl-org/sdl12-compat/blob/main/src/SDL12_compat.c#L1081-L1133

Where do you put your SDL2 dylib/framework along with your own binary?

Kreeblah:
I was wondering how it found the SDL 2.0 dylib.  That explains it.  I guess I'll need to figure out another solution here, then, if that's not recommended.

The ones I've been using are the ones built by Homebrew, and I'm on an M1 Mac, so they're in `/opt/homebrew/opt/sdl2/lib/` (`/opt/homebrew/opt/sdl2/` being a directory that gets symlinked to the directory where the current version is installed).  The `sdl12-compat` dylib ends up in `/opt/homebrew/opt/sdl12-compat/lib/`.

[...]

PR #231, opened 2022-10-04
Updated FindSDL2 to expose shared library location for packaging purposes

Kreeblah:
I'm not really sure that this is the right place to put this, but since this is the module specifically included with `sdl12-compat`, it seems like a possible place.

In https://github.com/libsdl-org/sdl12-compat/issues/217 I had an issue where `dylibbundler` was creating Mac app bundles that didn't work when `sdl12-compat` was used.  It turned out that copying the SDL2 shared library to the same directory as the other libraries in the app bundle resolved things for me.  However, that involved some fiddly stuff in my CMake build process for the project I was working on, and it seemed like it would be beneficial to move those to the FindSDL2 module and expose the shared library location in addition to the other variables that are returned from it.

This PR includes my changes to the FindSDL2 module, in case other people run into similar issues with packaging.  I tried to include changes for not just macOS but also Windows, OS/2, and more traditional *NIX OSes, though I only really have the ability to test this on macOS right now.  However, if other projects on other platforms create packages to bundle libraries with a binary distribution like I was doing, it could come in handy there.  Ideally, they'd move to use SDL2 directly, but that may not always be feasible.

So, I'm hoping to contribute this change for other folks to use for similar use cases.

### DIFF
--- a/cmake/modules/FindSDL2.cmake
+++ b/cmake/modules/FindSDL2.cmake
@@ -57,6 +57,7 @@ This module will set the following variables in your project:
 ::
 
   SDL2_LIBRARIES, the name of the library to link against
+  SDL2_SDL12_COMPAT_SHARED_LIBRARY_PATH, the location of the shared library loaded by sdl12-compat
   SDL2_INCLUDE_DIRS, where to find SDL.h
   SDL2_FOUND, if false, do not try to link to SDL2
   SDL2MAIN_FOUND, if false, do not try to link to SDL2main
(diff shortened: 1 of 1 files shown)


madebr:
sdl12-compat only uses `SDL2_INCLUDE_DIRS` from `FindSDL2.cmake`.
So making it create extra variables would complicate it for no added value.
There is even not going on any [`configure_file`](https://cmake.org/cmake/help/latest/command/configure_file.html) or similar.

`FindSDL2.cmake` is an sdl12-compat internal cmake module, so it should not be depended upon for packaging SDL2.
SDL2's cmake'ry handles this.

Kreeblah:
While I understand that, right now, it's just used for getting the include locations from SDL2 when building this library, I was concerned that including this change in a module in the SDL2 project directly would be, in some ways, a strange thing to do, since its only use case is for helping to package a different library.

If that's the preferred place for it, though, I'm happy to submit a PR there instead.

madebr:
Information about the location of libraries is available in the `SDL2Config.cmake`/`sdl2-config.cmake` files, installed by SDL2.
Since the last release, `SDL2::SDL2` contains the location of the SDL2 library (through the `IMPORTED_LOCATION` property).
The Macos framework binary release is an exception.

Kreeblah:
OK, I'll take a look at it and open an issue over there if it doesn't expose the info I need.

Thanks!

PR #267, opened 2022-10-17
quirks: Fixes for symbol-based quirks to force X11

sulix:
Make some changes to how the automatic quirks for apps which link against gl[x]ew and Cg functions work. In particular:
- They can be overridden by a user-set SDL_VIDEODRIVER env variable.
- They only print warnings if debug logging (SDL12COMPAT_DEBUG_LOGGING=1) is enabled, to match the other quirks.
- No message is printed if the user set SDL_VIDEODRIVER=x11 manually anyway.
- The messages are no-longer wrapped.

Fixes: 2a96bb35d756 ("quirks: Force X11 on Linux if certain symbols are in the process.")

In short, it now behaves more like the other quirks.

Note that there are some downsides to this:
- The messages are probably useful enough, even to people who don't typically enable debug logging.
- A number of setups will set SDL_VIDEODRIVER by default, so it won't be overridden. (And, without debug logging, will fail silently, as before.)
- Those messages are quite long, so wrapping is not without its merits.

On the whole, though, I think this is worthwhile. I've definitely managed to get several Cg games working under wayland using a [hacky LD_PRELOADed shim](https://gist.github.com/sulix/fe951daaeb4e51dfb329e5935e705cbf), which requires _some_ way of overriding the quirk here.

### DIFF
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -1307,6 +1307,7 @@ static void
 SDL12Compat_ApplyQuirks(void)
 {
     const char *exe_name = SDL12Compat_GetExeName();
+    const char *videodriver_env = SDL20_getenv("SDL_VIDEODRIVER");
     int i;
 
     if (*exe_name == '\0') {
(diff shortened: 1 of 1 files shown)


icculus:
I wonder how hard it would be to bind a socket, force `DISPLAY` to point at it, and speak just enough of the X11 protocol to make these apps survive extension lookup and glXGetProcAddress.  :)

icculus:
(for future reference: GitHub's "here's how to update this PR from the command line" instructions don't update the PR, they push right to main.  :/  )

Issue #279, opened 2022-11-24
Alpha Centauri - Artifacts around video playback


vv221:
Alpha Centauri used to be provided with a native engine developed by Loki, that can still be used with ./play.it (using the game build sold on GOG as a basis): https://www.dotslashplay.it/en/games/alpha-centauri

Video of secret projects are surrounded by some artifacts, see the following screenshot:
![smac](https://user-images.githubusercontent.com/10319639/203833328-f50e0eab-2c7c-4c8e-9d35-2cb7eb92d8d4.png)
With the real SDL 1.2 library, the parts of the screen surrounding the video playback are regular black bars, with no visible artifact.

The opening video (played when launching the game) is not affected.

The attached saved game is one turn before the completion of a secret project. Ending the turn triggers the video playback: [secret-project-video.sav.gz](https://github.com/libsdl-org/sdl12-compat/files/10094235/secret-project-video.sav.gz)

The file needs to be ungzipped, then copied into `~/.loki/smac/saves/`.

Since this would require access to a commercial game for investigating the issue, I would be OK to buy a gift code to someone willing to work on this issue but does not own a copy of the game yet.

I am of course willing to provide more information, but since I’m quite new to anything related to SDL I would need step-by-step instructions on how to debug the issue.

vv221:
More details on the Alpha Centauri build used to trigger this bug is shared in another issue: https://github.com/libsdl-org/sdl12-compat/issues/277#issuecomment-1327217654

sulix:
I can't reproduce this (at least with the opening video) with my copy of SMAC here. That's the full Loki version (with the 6.0a patch applied), not the ./play.it modified version, with some version of my [wrapper library](https://gist.github.com/sulix/09f059c468fcb58eefbca8ef46006ccb), which looks pretty similar to the one you mentioned in another of these threads.

vv221:
> I can't reproduce this (at least with the opening video) with my copy of SMAC here.

My bad, I forgot to add this precision: this does not happen with the opening video, only during gameplay when a secret project video is played.

vv221:
> That's the full Loki version (with the 6.0a patch applied), not the ./play.it modified version

[...]

PR #280, opened 2022-11-27
DisplayYUVOverlay: Don't update screen surface if clean

sulix:
Currently, if there are multiple SDL_DisplayYUVOverlay() calls in a row, we force an SDL_Flip() between them, in order to make sure that the screen is updated. However, this results in the non-YUV video surface also being updated, even if it otherwise wouldn't be.

Instead, copy the code from all of the other places where we force an update of the screen, which checks if the surface is dirty, and if not calls PresentScreen() directly without updating the VideoSurface. (After checking we're in the correct thread, of course...)

Note that it's probably worth pulling the "force screen update" code out into a separate helper function, rather than copying it everywhere.

This hopefully fixes video artefacts in Alpha Centauri (bug #279)

### DIFF
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -7474,6 +7474,7 @@ SDL_DisplayYUVOverlay(SDL12_Overlay *overlay12, SDL12_Rect *dstrect12)
     QueuedOverlayItem *overlay;
     SDL12_YUVData *hwdata;
     SDL_Renderer *renderer = NULL;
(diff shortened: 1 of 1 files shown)


vv221:
I should be able to give a try to this patch soon, I will share my feedback as soon as it’s done.

vv221:
I tested both current master: I get the artifacts reported in #279;
and this patch: I do not get any artifact.

So it seems good to me, it does get rid of the screen artifacts on my setup.

Issue #278, opened 2022-11-24
Alpha Centauri - Game is stuck after playing a video


vv221:
Alpha Centauri used to be provided with a native engine developed by Loki, that can still be used with ./play.it (using the game build sold on GOG as a basis): https://www.dotslashplay.it/en/games/alpha-centauri

After playing the video of a secret project, the game is stuck on a black screen (maybe the last frame of the video). In addition, these videos can no longer be skipped. With the real SDL 1.2 the video can be skipped by clicking, and after it is played in full we are back to the regular game screen.

Since this would require access to a commercial game for investigating the issue, I would be OK to buy a gift code to someone willing to work on this issue but does not own a copy of the game yet.

I am of course willing to provide more information, but since I’m quite new to anything related to SDL I would need step-by-step instructions on how to debug the issue.

vv221:
The intro video played when starting the game does not have this issue. It can be skipped the usual way.

vv221:
More details on the Alpha Centauri build used to trigger this bug is shared in another issue: https://github.com/libsdl-org/sdl12-compat/issues/277#issuecomment-1327217654

vv221:
For some reason this no longer triggers, despite not changing anything on my host system (no upgrade of anything related to SDL). If I find more details on what triggers the freeze I will share them.

sulix:
The inability to skip videos happens pretty sporadically for me, too. It mostly happens under Wayland, but even then, it's possible to skip the video if you're quick enough.

It looks like an issue with SDL_EVENTTHREAD: PR #281 seems to fix it here.

icculus:
Will close this when @vv221 has a chance to confirm the PR we merged solved this problem as well.

vv221:
While I did not get the game freeze again, I still am sometimes unable to skip videos if I do not click in the first couple seconds (for some reason this is not always the case). This includes both the intro video, and the secret project ones.

For the intro video this is an issue that I get with the real SDL 1.2 too. For the secret project ones I would need to run some extra tests, as I tend never to skip them during regular gameplay so it might have been a long standing issue with real SDL 1.2 that I failed to notice in the first place.

Since the limited ability to skip videos does not seem to be specific to sdl12-compat, and I did not manager to replicate the game getting stuck after video playback, maybe it is worth closing this issue and opening a new one specific to the skipping issue? Or maybe it’s even out of scope for this library, as it happens with real SDL 1.2 too (at least for the intro video)?

[...]

PR #281, opened 2022-11-27
Call PumpEvents in PresentScreen when SDL_EVENTTHREAD enabled

sulix:
At the moment, we "emulate" SDL_EVENTTHREAD by calling SDL_PumpEvents() in SDL_PeekEvent(). But some games don't even do this, so events aren't pumped at all, particularly while (e.g.) videos are playing, which is often just a tight SDL_DisplayYUVOverlay() loop.

If we SDL_PumpEvents() from within PresentScreen(), we should at least get events as long as the screen is being updated.

The complexity here is that SDL_PumpEvents() itself calls PresentScreen() to get a minimal framerate from things rendering directly to the front buffer. So we add a quick check to avoid SDL_PumpEvents() from entering infinite recursion by calling itself.

Should fix the difficulty with skipping videos in Alpha Centauri mentioned in #278.

### DIFF
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -6593,6 +6593,17 @@ PresentScreen(void)
         return;
     }
 
+    /* We don't actually implement an event thread in sdl12-compat, but some
+     * games will only call SDL_PeepEvents(), which doesn't otherwise pump
+     * events, and get stuck when they've consumed all the events.
+     *
+     * Just pumping the event loop here simulates an event thread well enough
+     * for most things.
+     */
+    if (EventThreadEnabled) {
+        SDL_PumpEvents();
+    }
+
     SDL20_RenderClear(renderer);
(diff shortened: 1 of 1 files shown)


vv221:
This one will be a bit tricky to test, because the inability to skip video seems to happen in two distinct ways:
- the intro video can sometimes only be skipped early, you have to watch it until the end if you did not skip it in the first couple seconds. This does not lead to anything game breaking, even if it can be annoying.
- #278 is about secret project videos being unskippable, and not letting the gameplay go on after the end of the video (the screen stay on the last video frame or a black screen, and the layered text at the end is never displayed).

Neither of these bugs seems to be systematic, so I will do multiple tests with your patch and report here once I am confident that it improves the situation for either of the symptoms.

sulix:
No worries. I'm pretty sure this fixes the unskippable video issue — I haven't been able to reproduce it since, and it makes sense that this could be the cause.

I never encountered the second issue, so who knows…

PR #293, opened 2023-05-27
Check for undefined TARGET_OS_*

jmroot:
Prevents some versions of clang from erroring out like this:
```
error: 'TARGET_OS_TV' is not defined, evaluates to 0 [-Werror,-Wundef-prefix=TARGET_OS_]
```


icculus:
We might need this fix in the real SDL headers too.

slouken:
Yup, I believe we do.

sezero:
Those aren't actually in real SDL-1.2 headers.

sezero:
Pushed https://github.com/libsdl-org/sdl12-compat/commit/c1eb2fecf9c5fc05e2efce0e2a135cd7474a4874 as a follow-up.

Issue #299, opened 2023-06-27
Advanced Strategic Command (asc) fails to build against sdl12-compat headers


smcv:
To reproduce:

* Debian testing/unstable (Debian 12 is probably also affected, I haven't tried)
* `sudo apt build-dep asc`
* `apt source asc`
* `cd asc-*/`
* `dpkg-buildpackage -b` (or compile it in whatever other way you prefer)
* `sudo apt install libsdl1.2-compat-dev` (this replaces "classic SDL 1.2", `libsdl1.2-dev`, with sdl12-compat headers)
* `dpkg-buildpackage -b` again

Expected result: both calls to `dpkg-buildpackage` succeed.

Actual result:

The first build (against classic SDL 1.2) works.

The second build fails:

```
/bin/bash ../../../../libtool  --tag=CXX   --mode=compile g++ -DHAVE_CONFIG_H -I. -I../../../..   -Wdate-time -D_FORTIFY_SOURCE=2 -I/usr/lib/x86_64-linux-gnu/wx/include/gtk3-unicode-3.2 -I/usr/include/wx-3.2 -D_FILE_OFFSET_BITS=64 -DWXUSINGDLL -D__WXGTK__ -I/usr/include/SDL -D_GNU_SOURCE=1 -D_REENTRANT  -g -O2 -ffile-prefix-map=/home/smcv/tmp/asc=. -fstack-protector-strong -Wformat -Werror=format-security -std=c++11 -Wno-sign-compare -D_UNIX_ -D_SDL_ -I/usr/include/freetype2 -I/usr/include/libpng16   -DSIZE_T_not_identical_to_INT -c -o sdlmm_pixelformat.lo sdlmm_pixelformat.cpp
In file included from sdlmm_display.h:25,
                 from sdlmm_display.cpp:25:
sdlmm_basesurface.h: In member function 'SDLmm::private_hwdata* SDLmm::BaseSurface::hwdata() const':
sdlmm_basesurface.h:142:66: error: invalid conversion from 'void*' to 'SDLmm::private_hwdata*' [-fpermissive]
  142 |     struct private_hwdata *hwdata() const { return GetSurface()->hwdata; }
      |                                                    ~~~~~~~~~~~~~~^~~~~~
      |                                                                  |
      |                                                                  void*
```

[...]

PR #300, opened 2023-06-27
include: Use `struct private_hwdata` incomplete type for API compat

smcv:
Classic SDL 1.2 headers traditionally had this member as a pointer to the incomplete type `struct private_hwdata`. For whatever reason (presumably a sense of completeness) the SDLmm C++ binding[1] that is bundled in Debian package 'asc' (Advanced Strategic Command) defines an accessor for it, which is declared as returning `struct private_hwdata *` and has the obvious trivial implementation[2].

Implicit conversion from `void *` to `struct private_hwdata *` is allowed in C, but not in C++, so sdl12-classic's simplification of changing this member from `struct private_hwdata *` to `void *` breaks the build for SDLmm.

Arguably it's a bug that SDLmm has this accessor, because this member is for internal use by SDL and its type has no public definition, so there's no legitimate reason for third-party code to call the accessor; but it does exist, so 100% source compatibility with classic SDL 1.2 would require it to have the same type it historically had.

[1] https://sourceforge.net/projects/sdlmm/  
[2] https://sourceforge.net/p/sdlmm/code/HEAD/tree/trunk/SDLmm/src/sdlmm_basesurface.h#l141

Resolves: https://github.com/libsdl-org/sdl12-compat/issues/299

---

If the sdl12-compat maintainers prefer to reject this PR as "SDLmm is wrong, sdl12-compat is already fine" then the change to fix compilation of asc should be as simple as deleting the `hwdata` accessor, which asc never calls anyway.

### DIFF
--- a/include/SDL/SDL_video.h
+++ b/include/SDL/SDL_video.h
@@ -89,7 +89,7 @@ typedef struct SDL_Surface
     Uint16 pitch;
     void *pixels;
     int offset;
-    void *hwdata;
+    struct private_hwdata *hwdata;
     SDL_Rect clip_rect;
     Uint32 unused1;
     Uint32 locked;


smcv:
> If the sdl12-compat maintainers prefer to reject this PR as "SDLmm is wrong, sdl12-compat is already fine" then the change to fix compilation of asc should be as simple as deleting the hwdata accessor, which asc never calls anyway.

Confirmed: deleting the accessor is enough to make SDLmm and asc compile successfully against sdl12-compat headers.

slouken:
There's no reason not to have complete source compatibility with SDL 1.2. I'll go ahead and merge this. Thanks!

Issue #305, opened 2023-07-18
makes SDL_perl regress when it (incorrectly) frees a former video surface


smcv:
Since trying to switch from classic SDL 1.2 to sdl12-compat in Debian, we've been seeing crashes in the test `perl t/core_video.t` in the test suite of [the SDL Perl bindings](https://metacpan.org/dist/SDL) (packaged as [libsdl-perl](https://tracker.debian.org/pkg/libsdl-perl)). On 32-bit platforms it seems to segfault repeatably. On 64-bit platforms it usually succeeds, but using valgrind or AddressSanitizer reveals that there is still a use-after-free, it's just not fatal for whatever reason.

I was able to cut down the failing test to this C code, which I think is equivalent to what the relevant part of `t/core_video.t` is doing:

```c
/* gcc -ot t.c `pkg-config --cflags --libs sdl`
 * SDL_VIDEODRIVER=dummy valgrind --leak-check=no ./t
 */
#include <SDL/SDL.h>

int main(void)
{
    SDL_Surface *one;
    SDL_Surface *two;

    one = SDL_SetVideoMode(640, 480, 32, SDL_SWSURFACE);
    two = SDL_SetVideoMode(100, 100, 16, SDL_SWSURFACE);

    /* I'm not sure what order the Perl code is really freeing them in, it's using automatic memory management */
    SDL_FreeSurface(two);
    SDL_FreeSurface(one);
    return 0;
}
```

With classic SDL 1.2 (1.2.15, from Debian 12) this succeeds, and `valgrind --leak-check=no` reports no memory errors.

With sdl12-compat (1.2.64 plus the fixes I contributed since then, from Debian unstable) this fails:

```
==17851== Invalid read of size 4
==17851==    at 0x4865EEC: SDL_FreeSurface (in /usr/lib/i386-linux-gnu/libSDL-1.2.so.1.2.64)
==17851==    by 0x1091FA: main (in /home/user/t)
==17851==  Address 0x4c9b680 is 56 bytes inside a block of size 60 free'd
==17851==    at 0x4841EA7: free (vg_replace_malloc.c:872)
==17851==    by 0x4F4BE17: ??? (in /usr/lib/i386-linux-gnu/libSDL2-2.0.so.0.2800.1)
==17851==    by 0x4F4C170: ??? (in /usr/lib/i386-linux-gnu/libSDL2-2.0.so.0.2800.1)
[...]

PR #306, opened 2023-07-20
Allocate video surface object statically as a global

smcv:
In #305, we have trouble with libsdl-perl making assumptions about the video surface never being freed or reallocated, and it being a no-op to "free" the video surface.

There is nothing that says a SDL 1.2 surface must always wrap the same SDL 2 surface, so we can maybe accommodate those assumptions by having the video surface be a special-cased global pointer that is never freed, and only reallocate its contents?

This seems to work with libsdl-perl's `core_video.t` test, but I haven't tested it extensively; consider it a proof-of-concept for now.

* Factor out Surface20to12InPlace
    
    This is a step towards making it possible to use the same SDL12_Surface
    for the video surface at all times, and only reallocating its underlying
    SDL 2 SDL_Surface.

* Factor out FreeSurfaceContents, and handle surface20 more defensively
    
    This is a step towards making it possible to use the same SDL12_Surface
    for the video surface at all times, and only reallocating its underlying
    SDL 2 SDL_Surface.

* Factor out enough of SDL_CreateRGBSurface to create surfaces in-place
    
    The part before we create the SDL 1.2 surface (creating the SDL 2.0
    surface) becomes CreateRGBSurface(), and the part after becomes
    Surface12SetMasks().

* Allocate video surface object statically as a global
    
    The SDL Perl bindings incorrectly call SDL_FreeSurface() on the result
    of functions that return a "borrowed" pointer to the video surface,
    namely SDL_SetVideoMode() and SDL_GetVideoSurface().
    (See https://github.com/PerlGameDev/SDL/issues/305)
    
    When we would previously have allocated or freed the video surface
    wrapper object, instead allocate or free its contents in-place.
    
    When checking whether the video surface exists, because we never destroy
    it, we must now also check whether its underlying SDL2 video surface
    exists.
    
    Resolves: https://github.com/libsdl-org/sdl12-compat/issues/305

### DIFF
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -5131,11 +5131,10 @@ SetPalette12ForMasks(SDL12_Surface *surface12, const Uint32 Rmask, const Uint32 
     }
 }
 
-DECLSPEC12 SDL12_Surface * SDLCALL
-SDL_CreateRGBSurface(Uint32 flags12, int width, int height, int depth, Uint32 Rmask, Uint32 Gmask, Uint32 Bmask, Uint32 Amask)
+static SDL_Surface *
+CreateRGBSurface(Uint32 flags12, int width, int height, int depth, Uint32 Rmask, Uint32 Gmask, Uint32 Bmask, Uint32 Amask)
 {
     SDL_Surface *surface20;
-    SDL12_Surface *surface12;
 
     /* SDL 1.2 checks this. */
     if ((width >= 16384) || (height >= 65536)) {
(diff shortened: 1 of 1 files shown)


smcv:
@slouken, @icculus: I think this is ready for at least a preliminary "is this a good direction?" review, but it isn't sufficiently tested to land it yet.

slouken:
This seems like a good direction to me.

smcv:
Rebased on main and fixed a missing NULL check. Briefly tested with:

* libsdl-perl: frozen-bubble, pangzero
* non-Perl: amoebax, armagetronad, enemylines7, powermanga

and seems to work as intended.

smcv:
I also contributed a fix for the incorrect free (PerlGameDev/SDL#306) which was applied to Debian's libsdl-perl package, but I haven't seen any sign of it being reviewed upstream. I also accidentally found PerlGameDev/SDL#307 while testing that change, and it hasn't had any response either, so PerlGameDev/SDL might be relatively dead.

slouken:
Okay, well let's merge this then.

Thanks!

Issue #310, opened 2023-08-06
SDL_LoadWAV() crashes with invalid files


zmanuel:
With native SDL 1.2, that used to return NULL and set an appropriate error.

To reproduce, simply call the loopwave test with a nonsense argument.

The fix is a simple standard early NULL check with return in SDL_LoadWAV_RW, I'll submit a PR momentarily.

We found this because our silly old game uses SDL_LoadWAV to just speculatively load optional sound files, and that crashes now: https://forums3.armagetronad.net/viewtopic.php?t=39657
A workaround is already in the release pipeline, so no hurry.

Looking around:
The similar SDL_LoadBMP(_RW) handles NULL arguments fine already.
I have not looked too deeply into SDL_SaveBMP. Calling it with both arguments NULL crashes, too, but because of the NULL surface argument; in native SDL1.2, a NULL surface and a valid file would already crash.

PR #311, opened 2023-08-06
Add null check to SDL_LoadWAV_RW to avoid crashes

zmanuel:
Native SDL 1.2 and 2.0 both handle SDL_LoadWAV(null,...), SDL_LoadWAV("",...) and SDL_LoadWAV(nonexistent_file,...) by returning NULL and setting an error; this check resores that behavior.

Fixes https://github.com/libsdl-org/sdl12-compat/issues/310

### DIFF
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -8454,6 +8454,10 @@ SDL_LoadWAV_RW(SDL12_RWops *rwops12, int freerwops12,
     SDL_RWops *rwops20 = RWops12to20(rwops12);
     SDL_AudioSpec *retval = NULL;
 
+    if (!rwops20) {
+        return NULL;
+    }
+
     *buf = NULL;
 
     /* SDL2's LoadWAV requires a seekable stream, but SDL 1.2 didn't,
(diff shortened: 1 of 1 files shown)


src/SDL12_compat.c
```diff
@@ -8454,6 +8454,10 @@ SDL_LoadWAV_RW(SDL12_RWops *rwops12, int freerwops12,
     SDL_RWops *rwops20 = RWops12to20(rwops12);
     SDL_AudioSpec *retval = NULL;
 
+    if (!rwops20) {
```
sezero, line 8457: I wonder: Should you move this to after `*buf = NULL;` below?

zmanuel: You're correct, that would be better. I had not considered that before. I'll change it.

What I also did not check there was the behavior of native SDL 1.2 with regards to buf; turns out that it is also left unitialized there. I don't know your policy there, so it's your call: Code that makes sense -> check after setting *buf, highest compatibility -> check before setting *buf. My take: I do not see a sensible way a caller might depend on the old SDL 1.2 behavior.

PR #321, opened 2023-09-14
Add a SDL12COMPAT_MAX_BPP hint, and use it to restrict Hyperspace Delivery Boy to 16bpp (fixes #317)

### LINKED ISSUE #317: Hyperspace Delivery Boy [LGP] colorkeys are broken on >16bpp displays
This looks like a game bug, not an sdl12-compat bug, but since it works with the bundled SDL 1.2, who knows.

[Hyperspace Delivery Boy](https://www.linuxgamepublishing.com/info.php?id=17)'s colour keys do not work at all under sdl12-compat, or under a freshly built SDL 1.2, but do work with the game's bundled version of SDL 1.2.13.
Screenshot of the HDB title screen under sdl12-compat, with everything surrounded by magenta boxes

The game is setting the colour key to 0xF800F8, but the actual colour of the key in the images is 0xFFFF (they're 16-bit). This would work, but the game appears to be using ``SDL_ConvertSurface`` to convert the surfaces to the screen surface's colour depth, before calling ``SDL_MapRGB(screen_format, 0xF8, 0x00, 0xF8)``, so if the screen surface is >16-bit, the colour key no longer matches.

Hacking the game to use 16-bit (by modifying ``SDL_SetVideoMode()``) fixes the issue, but it persists in using 32-bit by default, as it:
- Calls ``SDL_GetVideoInfo()`` to get the default video format.
- If the bpp of that format is >= 16, it calls ``SDL_SetVideoMode()`` with that bpp.
- Otherwise, it calls ``SDL_SetVideoMode()`` with 16bpp.

The bundled version of SDL 1.2, however, does seem to return a 32-bit format from ``SDL_GetVideoInfo()`` as well, so maybe there's something else going on.

My feeling is that the closest thing to a "correct" hack here is to add a quirk which overrides the bit depth to 16, but I think what's going on with the bundled SDL 1.2 is possibly a bit more subtle.

sulix:
Some games (e.g. Hyperspace Delivery Boy) only work on a 16-bit display,
but request the highest bit-depth available. Add a hint which makes all
queries for a mode (including the current mode, and the implicit format
chosen by providng SDL_SetVideoMode() a bpp of 0) report a bit depth
less than or equal to the value of SDL12COMPAT_FORCE_BPP.

We then add a quirk to force Hyperspace Delivery Boy to 16-bit, as it uses
the current video mode's bpp, but for 32-bit modes relies heavily on the
exact way format conversion was broken in some SDL 1.2 versions.

There are a few interesting "features" of this implementation:
- All video modes are now clamped to 32bit, which is the default value
  of the hint. This seems like something we'd probably want anyway, as
  many SDL 1.2 apps would have trouble with >32bit video modes anway
  (and we never properly supported them).
- This is our first proper integer hint, so add an SDL_GetHintInt()
  function.
- Split the BPP -> format conversion into its own helper function.
- This doesn't clamp the bit depth of user-created surfaces, or of the
  screen if the application specifically requests a bit depth. It only
  changes defaults and reported modes.

Regardless, this seems, at least to me, to be a better solution than trying
to emulate all of the ways different versions of SDL 1.2 could get format
conversion wrong. And, if you're bored, you can force games to run at 8bpp
and enjoy the horrible colour banding.

### DIFF
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -1274,6 +1274,10 @@ static QuirkEntryType quirks[] = {
     {"fillets", "SDL12COMPAT_ALLOW_SYSWM", "0"},
     {"fillets", "SDL12COMPAT_COMPATIBILITY_AUDIOCVT", "1"},
 
+    /* Hyperspace Delivery Boy relies on the exact imprecision of the format conversion in some 
+       earlier versions of SDL 1.2. It also recommends 16-bit in the README, so force it. */
+    {"hdb", "SDL12COMPAT_MAX_BPP", "16"},
+
     /* Mark of the Ninja doesn't work with OpenGL scaling */
     {"ninja-bin32", "SDL12COMPAT_OPENGL_SCALING", "0"},
     {"ninja-bin64", "SDL12COMPAT_OPENGL_SCALING", "0"},
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -1396,6 +1396,18 @@ SDL12Compat_GetHintFloat(const char *name, float default_value)
     return (float) SDL20_atof(val);
 }
 
+static int
+SDL12Compat_GetHintInt(const char *name, int default_value)
+{
+    const char *val = SDL12Compat_GetHint(name);
+
+    if (!val) {
+        return default_value;
+    }
+
+    return SDL20_atoi(val);
+}
+
 static void
 SDL12Compat_ApplyQuirks(SDL_bool force_x11)
 {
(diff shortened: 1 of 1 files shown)


sezero:
`SDL_BITSPERPIXEL(mode.format) > max_bpp` triggers `-Wsign-compare`
If you change max_bpp to unsigned and BPPToPixelFormat to accept unsigned
all is good, like
```diff
diff --git a/src/SDL12_compat.c b/src/SDL12_compat.c
index f3ecd3b..f089a69 100644
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -2049 +2049 @@ static SDL_PixelFormatEnum
-BPPToPixelFormat(int bpp)
+BPPToPixelFormat(unsigned bpp)
@@ -2247 +2247 @@ Init12VidModes(void)
-    const int max_bpp = SDL12Compat_GetHintInt("SDL12COMPAT_MAX_BPP", 32);
+    const unsigned max_bpp = SDL12Compat_GetHintInt("SDL12COMPAT_MAX_BPP", 32);
@@ -2416 +2416 @@ Init12Video(void)
-    const int max_bpp = SDL12Compat_GetHintInt("SDL12COMPAT_MAX_BPP", 32);
+    const unsigned max_bpp = SDL12Compat_GetHintInt("SDL12COMPAT_MAX_BPP", 32);
```

sulix:
> `SDL_BITSPERPIXEL(mode.format) > max_bpp` triggers `-Wsign-compare` If you change max_bpp to unsigned and BPPToPixelFormat to accept unsigned all is good[…]

Whoops — nice catch, thanks!

PR #325, opened 2023-12-31
cdrom: fix 'no track 1' edge case

twolife:
unless I'm missing something obvious, `c` is never equal to zero


icculus:
Yeah, you're right, this is a leftover from when the tracks started at zero, I think.

sezero:
Issue was introduced in https://github.com/libsdl-org/sdl12-compat/commit/ed299ba1772ee52c58b7817f3997101b5c79ff14
(original check was `if (!rw && c > 1)` )

PR #333, opened 2024-05-15
quirks: Add SDL12COMPAT_COMPATIBILITY_AUDIOCVT=1 to "heroes3.dynamic"

twolife:
Don't ask my why this is needed, but without it the game just freeze/crash after the intro video:

```
INFO: This app is looking for CD-ROM drives, but no path was specified
INFO: Set the SDL12COMPAT_FAKE_CDROM_PATH environment variable to a directory
INFO: of MP3 files named trackXX.mp3 where XX is a track number in two digits
INFO: from 01 to 99
double free or corruption (!prev)

BUG!  Exception triggered, cleaning up.
Heroes of Might & Magic III 1.3.1a
Built with glibc-2.1 on x86
Stack dump:
{
```


slouken:
Merged, thanks!

PR #349, opened 2025-03-06
video: Reflect SDL12_RESIZABLE in SDL12_Surface->flags

neuschaefer:
Code depends on being able to see the RESIZABLE flag in video surfaces. For example, the Perl module SDLx::App will throw an error if it tries to resize a surface but doesn't find SDL_RESIZABLE in its flags:

<https://metacpan.org/release/FROGGS/SDL-2.548/source/lib/SDLx/App.pm#L143>

To improve compatibility with such code, copy the flag to the VideoSurface12 created in SDL_SetVideoMode.


slouken:
Merged, thanks!

PR #355, opened 2025-04-06
Revert "Rename sdl12_compat.pc to sdl.pc"

Conan-Kudo:
This is unnecessary, given that everyone uses `pkgconf(1)`, which fully supports the `Provides:` stanza and expresses it in dependencies.

This reverts commit f10920239783dc09bd22f09bad69585cfce8a28d.


sezero:
> This is unnecessary

Does it really break anything though?

Conan-Kudo:
Please see the discussion in https://github.com/libsdl-org/sdl2-compat/pull/451#event-17173710855

Issue #339, opened 2024-06-30
issues affecting Loki port of Sim City 3000


twolife:
I know that this is borderline "out-of-scope", as that game is an ancient SDL 1.1 game unplayable with a modern glibc without using some nasty hack like https://github.com/twolife/sc3u-nptl, but here it is, I'm running this game on a modern stack (including sdl12-compat) and it almost works perfectly :-) 

The first issue is that the intro movie is flickering. I'm surprised as I saw issue #49 that was resolved, but the intro movie in SC3U is still affected.

<details>
  <summary>Second issue is that the game segfault on exit</summary>

  ```
BUG! (Segmentation Fault)  Going down hard...
SimCity 3000 Unlimited 2.0.955a
Built with glibc-2.1 on x86
Stack dump:
{
	0x83fc566
	0xf66693b7
	0xf6669e01
	0xf664dba4
	0xf66b5a49
	0xf7f38e41
	0xf7936086
}
Please send a full bug report,
along with the contents of autosave to: support@lokigames.com
Unable to execute loki_qagent - exiting
  ```
</details>

Those are very minor issues, but this is not happening with a real SDL-1.2, so i suppose it's a bug #sorry

Anyway, this is at least documented now.

icculus:
> #sorry

We definitely fixed other issues in other Loki binaries, so this is 100% in scope, fwiw.

terinjokes:
```
Thread 1 "sc3u.dynamic" received signal SIGSEGV, Segmentation fault.
0x00000000 in ?? ()
(gdb) bt
#0  0x00000000 in ?? ()
#1  0xf7df0cb3 in SDL_DestroyMutex (a=0x865d4d0) at /var/tmp/portage/media-libs/libsdl-1.2.68/work/sdl12-compat-release-1.2.68/src/SDL20_syms.h:179
#2  0x083fc566 in _._18cRZCriticalSection ()
#3  0xf6c223b7 in ?? () from /home/terin/SC3U/lib/libAudio.so
#4  0xf6c22e01 in ?? () from /home/terin/SC3U/lib/libAudio.so
#5  0xf6c06ba4 in ?? () from /home/terin/SC3U/lib/libAudio.so
#6  0xf6c6ea49 in _fini () from /home/terin/SC3U/lib/libAudio.so
#7  0xf7fcc2e9 in _dl_fini () at dl-fini.c:114
#8  0xf799ff4a in ?? () from /usr/lib/libc.so.6
[...]

PR #360, opened 2025-07-22
Add a new quirk 'SDL12COMPAT_NO_UNLOAD', usable/needed by SC3U

twolife:
This fixes the segfault when exiting the game described in #339

I don't really know if the path I choose to fix this is the right one, as I don't have a complete technical understanding of what's happening (but I guess the difference between a real SDL-1.2 and sdl12-compat is that fact that we explicitly calls `dlclose()` here)


icculus:
I think this is probably the correct fix, honestly. The other alternative is to _exit() at this point so everything goes away immediately without crashing, but that feels risky for like 14 other reasons, and we'd still need a new quirk in place anyhow.

PR #363, opened 2025-07-31
Hint: Add SDL12COMPAT_WINDOWED_MODE_LIST hint, to make SDL_ListModes() work in windowed mode.

sulix:
Add a new boolean hint, which makes SDL_ListModes() always return a mode list, even if the flags passed in are for windowed mode. By default, we return -1 in this case, to indicate that any mode is supported.

However, some older titles pass 0 for flags here even if they're intending to run in fullscreen, and if -1 is returned, attempt to populate their own mode list (and select a mode) themselves. For example, Civilization: Call To Power will add the current desktop mode (if it can glean it from X11, which doesn't work on Wayland, and gives an unusuable combined size for multi-monitor setups even on X11), along with the hardcoded list of 1024×768, 800×600, and 640×480).

By always returning sdl12-compat's list of modes (some of which may be "fake"), we get a much more modern list of supported modes, which allows the game to comfortably run at higher resolutions.

### DIFF
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -5662,6 +5662,7 @@ SDL_ListModes(const SDL12_PixelFormat *format12, Uint32 flags)
     VideoModeList *best_modes = NULL;
     Uint32 bpp;
     int i;
+    SDL_bool windowed_mode_list = SDL12Compat_GetHintBoolean("SDL12COMPAT_WINDOWED_MODE_LIST", SDL_FALSE);
 
     if (!SDL20_WasInit(SDL_INIT_VIDEO)) {
         SDL20_SetError("Video subsystem not initialized");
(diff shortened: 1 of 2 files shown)


src/SDL12_compat.c
```diff
@@ -5677,7 +5678,7 @@ SDL_ListModes(const SDL12_PixelFormat *format12, Uint32 flags)
         return (SDL12_Rect **) -1;  /* 1.2's dummy driver always returns -1, and it's useful to special-case that. */
     }
 
-    if (!(flags & SDL12_FULLSCREEN)) {
+    if (!windowed_mode_list && !(flags & SDL12_FULLSCREEN)) {
```
slouken, line 5681:
```suggestion
    if (!(flags & SDL12_FULLSCREEN) && !windowed_mode_list) {
```
Very minor tweak, we only check windowed_mode_list when we know that we're looking for windowed modes.

PR #366, opened 2025-09-30
Treat DEBUG_INVOCATION=1 as equivalent to SDL12COMPAT_DEBUG_LOGGING=1

smcv:
In the same spirit as https://github.com/libsdl-org/SDL/issues/12275, if we enable a reasonable level of debug output whenever this variable is set, it becomes a convenient way for users to request a more verbose (but not overwhelming) amount of output. SDL 3 and GLib already do this.

Explicitly setting SDL12COMPAT_DEBUG_LOGGING=0 can override this, if it becomes necessary for whatever reason.

### DIFF
--- a/src/SDL12_compat.c
+++ b/src/SDL12_compat.c
@@ -1222,6 +1222,8 @@ static const char *SDL12COMPAT_GetEnvAtStartup(const char *name)
 static SDL_bool SDL12COMPAT_CheckDebugLogging(void)
 {
     const char *value = SDL12COMPAT_GetEnvAtStartup("SDL12COMPAT_DEBUG_LOGGING");
+    if (value == NULL)
+        value = SDL12COMPAT_GetEnvAtStartup("DEBUG_INVOCATION");
     return ((value != NULL) && SDL12COMPAT_strequal(value, "1")) ? SDL_TRUE : SDL_FALSE;
 }
 


smcv:
With `LD_LIBRARY_PATH` set to point to this and the equivalent sdl2-compat build, setting `DEBUG_INVOCATION=1` is a convenient way to say "debug everything, please":

```
$ DEBUG_INVOCATION=1 amoebax
sdl2-compat 2.32.57, built on Sep 30 2025 at 13:03:05, talking to SDL3 3.3.0
sdl2-compat: This app appears to be named: amoebax
SDL_DYNAMIC_API: Not overridden
sdl12-compat 1.2.69, built on Sep 30 2025 at 13:03:52, talking to SDL2 2.32.57
sdl12-compat: This app appears to be named: amoebax
INFO: App name: SDL Application
INFO: App version: <unspecified>
INFO: App ID: <unspecified>
INFO: SDL revision: release-3.2.6-1294-ged2cdceb66 (Debian 3.3~git20250925~ed2cdce+ds-1)
INFO: SDL_DEBUG: chose video backend 'wayland'
(amoebax:276210): GLib-GIO-DEBUG: 13:05:40.606: _g_io_module_get_default: Found default implementation dconf (DConfSettingsBackend) for ?gsettings-backend?
...
```

slouken:
Merged, thanks!

We're considering making a new sdl12-compat release. Do you have downstream patches that make sense to bring in? Do you have any preference whether we do this sooner or later?

smcv:
> Do you have any preference whether we do this sooner or later?

I was actually just thinking "it's been a while since the last release, maybe it's time for a new one" so this makes sense. Are you planning to do a release-candidate, or go straight to a stable release? If the latter, it's probably best if I package a git snapshot in experimental first (like I've been doing for sdl2-compat and libsdl3) so I can check for any nasty surprises.

> Do you have downstream patches that make sense to bring in?

No, only a cherry-pick from upstream. I generally try to upstream anything that makes sense, the only patches we keep as downstream are the ones that are really Debian-specific (multiarch-related changes that only work because we know we're installing into `/usr`, or changes related to Debian policies on licensing or avoiding pre-generated files).

slouken:
> I was actually just thinking "it's been a while since the last release, maybe it's time for a new one" so this makes sense. Are you planning to do a release-candidate, or go straight to a stable release? If the latter, it's probably best if I package a git snapshot in experimental first (like I've been doing for sdl2-compat and libsdl3) so I can check for any nasty surprises.

We'll probably do a release candidate, but I don't expect much feedback, so an experimental Debian release would be great.

smcv:
> probably best if I package a git snapshot in experimental first

Done, no obvious issues (but I'll see what the autobuilders and automated test infrastructure think of it).

PR #374, opened 2025-10-26
Added GitHub release workflow

slouken:


slouken:
I can't run the workflow without it existing in the main branch, so I'll merge this PR and iterate afterwards.

PR #377, opened 2025-10-30
test: Optionally install manual tests for "as-installed" testing

smcv:
* test: Optionally install manual tests for "as-installed" testing
    
    This provides a convenient way to install manual tests onto a
    test system. For example in Debian these are packaged into the
    libsdl1.2-compat-tests binary package, from which they can be run with
    commands like:
    
        cd /usr/libexec/installed-tests/SDL12_compat
        ./testoverlay2
    
    (The path used was historically
    /usr/libexec/installed-tests/libsdl1.2-compat, but this version uses a
    path that is more consistent with SDL2_compat.)

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -19,6 +19,11 @@ if(NOT TARGET SDL::SDLmain)
     endif()
 endif()
 
+option(SDL12COMPAT_INSTALL_TESTS "Install manual tests into libexecdir" OFF)
+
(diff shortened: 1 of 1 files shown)


smcv:
Testing this in a Debian package build, but I think it's right.

smcv:
> Testing this in a Debian package build, but I think it's right.

Yes, it works well.

slouken:
Merged, thanks!

PR #378, opened 2025-10-30
Install sdl.m4 to target dir instead of full dir

Sackzement:
When installing sdl12-compat to a custom directory, it still tries to install the file `sdl.m4` into the root directory `/usr/local/share/aclocal`.

Without privileges it fails:

```
CMake Error at sdl12-compat/cmake_install.cmake:119 (file):
  file cannot create directory: /usr/local/share/aclocal.  Maybe need
  administrative privileges.
```

This commit changes the install path of `sdl.m4` to the custom one.


madebr:
Indeed! Thank you for noticing and fixing!

Sackzement:
You're Welcome!

Issue #244, opened 2022-10-12
freedroid: Segfault after intro, etc


smcv:
Prerequisites:

* Debian testing (Debian 12 alpha)
* Video: GNOME 43 in Wayland mode (with Mesa 22.2.0 on AMD Vega, if it matters)
* Audio: Pipewire 0.3.59, with `pipewire-pulse` emulating PulseAudio
* `apt install freedroid` (Debian package version `1.0.2+cvs040112-7`) (not to be confused with 3D `freedroidrpg`, which is different)
* Some relevant libraries:
    * `libsdl1.2-compat` either 1.2.58-1 (packaged in Debian experimental) or commit 63e4393 (locally built)
    * `libsdl2-2.0-0` version `2.24.1+dfsg-1`

To reproduce:

* `freedroid`
* `LD_LIBRARY_PATH=.../sdl12-compat/_build freedroid`
* `SDL_VIDEODRIVER=wayland LD_LIBRARY_PATH=.../sdl12-compat/_build freedroid`

Each time, click once to pass the Briefing screen, then a second time to pass the description of your robot.

Expected result: all work

Actual result: with sdl12-compat, it crashes.

```
Video info summary from SDL:
----------------------------------------------------------------------
Is it possible to create hardware surfaces: no
Is there a window manager available: yes
Are hardware to hardware blits accelerated: no
Are hardware to hardware colorkey blits accelerated: no
Are hardware to hardware alpha blits accelerated: no
Are software to hardware blits accelerated: no
Are software to hardware colorkey blits accelerated: no
Are software to hardware alpha blits accelerated: no
Are color fills accelerated: no
Total amount of video memory in Kilobytes: 262144
Pixel format of the video device: bpp = 32, bytes/pixel = 4
Video Driver Name: wayland
----------------------------------------------------------------------
Found new graphics-theme: classic 
Found new graphics-theme: lanzz 
Found new graphics-theme: para90 
Game starts using theme: classic
Segmentation fault (core dumped)
```

```
#0  SDL_ListRemove (head=0x55b900e608a8, ent=ent@entry=0x55b9a5f4dce0) at ./src/SDL_list.c:70
#1  0x00007f49e83801eb in SDL_InvalidateMap (map=0x55b9a5f4dce0) at ./src/video/SDL_pixels.c:1052
[...]

PR #394, opened 2026-05-01
SDL_UpperBlit: fix an oversight in commit 61e247df for issue #244.

sezero:
SDL_UpperBlit: fix an oversight in commit 61e247df for issue #244.

Otherwise NULL pointer dereference can happen in PrepBlitDestRect()


sezero:
P.S.: Did Dungeons of Dredmor survive w/o this patch? (Re: commit: 6cad845ed)

icculus:
Dredmor has been working fine for a long time with sdl12-compat, so I guess it didn't hit this specific problem.

